PR #2, opened 2025-08-14
Add skip TLS certicate verification feature and error log enhancement

ohtanim:
## Summary
In actual production deployment work, we frequently encounter issues with SSL/TLS connections. In such cases, we forcibly establish SSL connections by skipping TLS certificate verification to isolate the problem. With the curl command, this can be achieved using the `--insecure` or `-k` option. However, our Client does not have such a feature.

This pull request introduces a new environment variable, as shown below, which enables this behavior when set appropriately. Additionally, it adds logging for HTTP-layer errors during REST API calls to facilitate easier root cause analysis.

Environment variable to skip TLS certificates verifications.
```bash
export DANGER_TLS_SKIP_CERT_VERIFY=1
```
or
```bash
export DANGER_TLS_SKIP_CERT_VERIFY=true
```


IceKhan13 (approved):
Convenient feature!

PR #18, opened 2025-11-06
direct-access: Change retry policy and strategy for HTTP429

ohtanim:
Add new feature `"iqp_retry_policy"` code to allow to enable retry behavior consistent with the IBM Quantum Platform when execution lanes are full. Rather than returning an HTTP 429 error, the client library automatically retries job submission.

This PR also contains:
- Fix `cargo clippy` error found in qiskit-runtime-service code
- Fix `cargo fmt `error in some place

### DIFF
--- a/dependencies/direct_access_client/README.md
+++ b/dependencies/direct_access_client/README.md
@@ -31,7 +31,7 @@ This Rust API client provides all functionalities provided by Direct Access API,
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


ohtanim:
@utzb @IceKhan13 This change is urgent required for upcoming QDC/SC preparations by our client. I would appreciate if you could approve as soon as possible. Thanks,

PR #38, opened 2026-02-12
[FIX] PASQAL - fix  task status showing batch status in stead of job status

awennersteen:
This meant that the status switched to running much to early when running emulation jobs.

### DIFF
--- a/src/pasqal/cloud.rs
+++ b/src/pasqal/cloud.rs
@@ -336,9 +336,7 @@ impl QuantumResource for PasqalCloud {
                         JobStatus::Canceling => TaskStatus::Cancelled,
                         JobStatus::Done => TaskStatus::Completed,
                         JobStatus::Canceled => TaskStatus::Cancelled,
-                        JobStatus::TimedOut => TaskStatus::Failed,
                         JobStatus::Error => TaskStatus::Failed,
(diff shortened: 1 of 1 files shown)


src/pasqal/cloud.rs
```diff
@@ -132,18 +132,24 @@ impl QuantumResource for PasqalCloud {
...
-                    BatchStatus::Canceled => TaskStatus::Cancelled,
-                    BatchStatus::TimedOut => TaskStatus::Failed,
-                    BatchStatus::Error => TaskStatus::Failed,
-                    BatchStatus::Paused => TaskStatus::Queued,
-                };
-                return Ok(status);
-            }
+            // Assuming a single job per batch for now,
+            // Will need to be updated if multiple jobs per batch are supported in the future
+            Ok(batch) => match self.api_client.get_job(&batch.data.job_ids[0]).await {
```
MatthieuMoreau0: This means we make two calls to pasqal cloud APIs each time we call `task_status`. Wouldn't it be best if we store the job_id as task_id instead so that we can get the job directly without the extra step to retrieve the job ID?

MatthieuMoreau0 (approved):
Tested locally 👍  One cleanup comment

PR #40, opened 2026-02-13
[REFACTOR] - Pasqal move auth to client and expose auth errors

awennersteen:
## Summary
Move Pasqal Cloud authentication and token refresh handling from QRMI into` pasqal_cloud_api` so QRMI no longer owns auth state/refresh logic.

### What Changed
- Moved auth/token lifecycle into pasqal_cloud_api::Client:
- Raise error when trying to refresh but credentials are missing (AuthError::MissingCredentialsForRefresh)
- I tried to make this futureproof for oauth2 refresh token - I think that should be easily developed

- we now
  - configure a client via ClientBuilder
  - call client API methods directly
  - stop implementing auth refresh and setup  logic in QRMI
  
  
  Fixes
  
  - #57  
  
  - #58 
  
  
  
  In the latter case wrong auth creds now gives:
  
  
 ```bash
[root@login /]# tail -f /data/job_122.out 
Traceback (most recent call last):
  File "/shared/qrmi/examples/pulser_backend/pasqal/pulser_backend.py", line 26, in <module>
    service = QRMIService()
              ^^^^^^^^^^^^^
  File "/shared/qrmi/python/qrmi/pulser_backend/service.py", line 32, in __init__
    raise RuntimeError(plugin_error)
RuntimeError: Failed to get device: Token request failed: 403 Forbidden {"error":"invalid_grant","error_description":"Wrong email or password."}
```

### DIFF
--- a/src/pasqal/cloud.rs
+++ b/src/pasqal/cloud.rs
@@ -258,11 +258,13 @@ impl QuantumResource for PasqalCloud {
         let device_type = match self.backend_name.parse::<DeviceType>() {
             Ok(dt) => dt,
             Err(_) => {
-                let valid_devices = ["FRESNEL",
+                let valid_devices = [
+                    "FRESNEL",
                     "FRESNEL_CAN1",
                     "EMU_MPS",
                     "EMU_FREE",
-                    "EMU_FRESNEL"];
+                    "EMU_FRESNEL",
+                ];
                 let err = format!(
                     "Device '{}' is invalid. Valid devices: {}",
                     self.backend_name,
@@ -302,11 +304,13 @@ impl QuantumResource for PasqalCloud {
             let device_type = match self.backend_name.parse::<DeviceType>() {
                 Ok(dt) => dt,
                 Err(_) => {
-                    let valid_devices = ["FRESNEL",
+                    let valid_devices = [
+                        "FRESNEL",
                         "FRESNEL_CAN1",
                         "EMU_MPS",
                         "EMU_FREE",
-                        "EMU_FRESNEL"];
+                        "EMU_FRESNEL",
+                    ];
                     let err = format!(
                         "Device '{}' is invalid. Valid devices: {}",
                         self.backend_name,
@@ -381,11 +385,13 @@ impl QuantumResource for PasqalCloud {
         let device_type = match self.backend_name.parse::<DeviceType>() {
             Ok(dt) => dt,
             Err(_) => {
-                let valid_devices = ["FRESNEL",
+                let valid_devices = [
+                    "FRESNEL",
                     "FRESNEL_CAN1",
                     "EMU_MPS",
                     "EMU_FREE",
-                    "EMU_FRESNEL"];
+                    "EMU_FRESNEL",
+                ];
                 let err = format!(
                     "Device '{}' is invalid. Valid devices: {}",
                     self.backend_name,

### CODE REVIEW

File: python/qrmi/pulser_backend/service.py [hunk 1]
```diff
@@ -25,6 +25,12 @@ class QRMIService:
     """Class for interacting with the QRMI resources"""
 
     def __init__(self):
+        # If resource acquisition failed in QRMI plugin,
+        # the plugin may expose the error reason via environment variable.
+        plugin_error = os.environ.get("QRMI_PLUGIN_ERROR")
+        if plugin_error is not None:
+            raise RuntimeError(plugin_error)
```
awennersteen: @ohtanim Could you confirm that this is the approach you meant we could take to get the auth errors exposed in this comment:

https://github.com/qiskit-community/qrmi/pull/33#issuecomment-3930938467

Much appreciated

dependencies/pasqal_cloud_client/src/client.rs
```diff
@@ -344,12 +420,42 @@ impl ClientBuilder {
...
             token,
             project_id,
+            auth_endpoint: String::new(),
+            username: None,
+            password: None,
         }
     }
 
+    /// Construct a new [`ClientBuilder`] with project id.
+    pub fn for_project(project_id: String) -> Self {
```
MatthieuMoreau0: It looks like this should become the basic interface for instantiating a ClientBuilder. Have you considered updating the `new` function to take only the project ID as argument? It would be a breaking change of course, but it's not used outside the QRMI so we are free to perform the changes in the same MR so I think it's acceptable

src/pasqal/tests/cloud.rs
```diff
@@ -47,30 +48,51 @@ async fn is_accessible_attempts_authentication() {
...
+}
+
+#[test]
+fn read_qrmi_config_env_value_handles_empty_and_missing_environment_key() {
+    // This test verifies that `read_qrmi_config_env_value_from_content()` correctly
+    // handles cases where the "environment" key is missing or empty for a resource.
+    let content = r#"{
+        "resources": [
+            {"name":"EMU_FREE","type":"pasqal-cloud","environment":{}},
+            {"name":"EMU_OTHER","type":"pasqal-cloud"}
```
MatthieuMoreau0: rm as per the discussion on the other MR

src/pasqal/cloud.rs
```diff
@@ -144,16 +139,22 @@ fn read_qrmi_config_env_value_from_content(
...
+        .ok()
+        .filter(|v| !v.trim().is_empty())
+        .or(cfg.username.clone().filter(|v| !v.trim().is_empty()));
+    let password = env::var("PASQAL_PASSWORD")
+        .ok()
+        .filter(|v| !v.trim().is_empty())
+        .or(cfg.password.clone().filter(|v| !v.trim().is_empty()));
+    (username, password)
+}
+
```
awennersteen: I don't like this function name

src/pasqal/cloud.rs
```diff
@@ -33,6 +32,8 @@ struct PasqalConfig {
     auth_endpoint: Option<String>,
 }
 
+const DEFAULT_PASQAL_CLOUD_AUTH_ENDPOINT: &str = "authenticate.pasqal.cloud/oauth/token";
```
awennersteen: hm

src/pasqal/cloud.rs
```diff
@@ -163,15 +165,22 @@ impl PasqalCloud {
...
+    /// * `~/.pasqal/config`: Optional fallback for `username`, `password`, `token`, `project_id`, `auth_endpoint`
     pub fn new(backend_name: &str) -> Result<Self> {
         debug!(
             "Initializing PasqalCloud QRMI for backend '{}'",
             backend_name
         );
+
         let cfg = read_pasqal_config(backend_name)?;
 
+        // Project ID
```
awennersteen: I added some comments as this `new` function is becoming very large to be able to read it better locally. 
I personally prefer this over making several small functions

Issue #36, opened 2026-02-11
[DRAFT] EPIC: Unit Tests and Continuous Integration


OkuyanBoga:
### What problem are you trying to solve?

QRMI library requires a systematic, cross-language test strategy.

### Describe the solution you’d like

Establish a complete unit/integration testing and CI strategy for QRMI.

#### Test cases to discuss and implement

1. **API tests against a dummy provider**
   - Use deterministic mock provider(s) for tests.

2. **Provider connectivity checks**
   - Add lightweight provider ping/connectivity tests (non-destructive).

3. **Cross-language parity checks**
   - Verify core behaviour across C, Rust, and Python interfaces.

4. **Vendor-specific function tests**
   - Validate vendor-specific code paths with given specifics such as topology-aware test design.

5. **Basic functionality lifecycle**
   - Validate accessibility, acquire/release, task start/stop, status/result, metadata/target flows.

6. **Docs/tutorial validation**
   - Ensure example/tutorial code remains runnable via CI tests.

#### CI design
- Github Actions & Workflow:
  - #46 
  - #41 
- Github Docs
- PR pipeline:
  - fast unit + API contract tests,
  - cross-language checks.
- Scheduled/nightly pipeline:
  - expanded integration/provider checks.
- Store detailed artifacts/logs for debugging failures.
- Optional quality checks (coverage thresholds, lint/type checks).

### Documentation, Adoption, Migration Strategy

- Introduce tests in phases (core first, then vendor specifics).
- Document local test workflows for each language binding.
- Keep CI jobs clearly named and mapped to subsystem ownership.

### Sub-issues

- [ ] Dummy-provider API contract test suite
- [ ] Provider connectivity tests
- [ ] C/Rust/Python tests
- [ ] Vendor-specific behaviour tests
- [ ] Core lifecycle tests
- [ ] Docs/tutorial tests
- [ ] CI workflow integration and reporting

### Acceptance Criteria

- [ ] PRs are validated by deterministic core tests.
- [ ] Nightly/scheduled jobs run extended integration checks.
- [ ] C/Rust/Python interfaces remain behaviourally consistent.
- [ ] Tutorials/examples are continuously validated.

PR #41, opened 2026-02-13
Add build github action

cclaudio:
Partially fix https://github.com/qiskit-community/qrmi/issues/36

Add GHA to build, lint, format check and unittest the code.

### DIFF
--- a/.github/workflows/unittest-rust.yml
+++ b/.github/workflows/unittest-rust.yml
@@ -1,0 +1,46 @@
+---
+# This reusable workflow runs Rust unittests on the libqrmi, binaries and dependencies.
+#
+# Workflow inputs:
+# - None
+#
+# Workflow artifacts produced:
+# - None
+#
+# Workflow dependencies:
+# - None. It does not depend on any other workflow.
+name: Unittest QRMI Rust code
+on:
+  workflow_call:
+env:
+  CARGO_TERM_COLOR: always
+jobs:
+  unittest:
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v6
+
+    - name: Setup Rust toolchain
+      uses: actions-rust-lang/setup-rust-toolchain@v1
+
+    - name: Unittest libqrmi.so (includes examples)
+      run: cargo test --all-targets --release
+
+    - name: Unittest task_runner
+      run: cargo test --bin task_runner --release --features="build-binary"
+
+    - name: Unittest stubgen
+      run: cargo test --bin stubgen --release --features="pyo3"
+
+    - name: Unittest direct_access_client
+      working-directory: dependencies/direct_access_client
+      run: cargo test --release
+
+    - name: Unittest pasqal_cloud_client
+      working-directory: dependencies/pasqal_cloud_client
+      run: cargo test --release
+
+    - name: Unittest qiskit_runtime_client
+      working-directory: dependencies/qiskit_runtime_client
+      run: cargo test --release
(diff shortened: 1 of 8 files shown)


ohtanim (changes requested):
@cclaudio Could you please revert all the changes except for your GitHub Actions workflow? 

As I told before, I believe each code owner should fix their own issues. Each developer needs to learn from the errors and try to avoid making the same mistakes in future code changes.

Thanks,

cclaudio:
@ohtanim I think both approaches have pros and cons. We can discuss them later, if you want.

For now, let me remove all patches from the PR, except the one that adds the github action. I am also adding the `continue-on-error: true` to all steps that are currently failing, I think that will help us understanding why they are failing.

cclaudio:
We mentioned `code owners` multiple times in this PR, however, QRMI does not seem to have a `CODEOWNERS` file yet.  I just created the issue https://github.com/qiskit-community/qrmi/issues/42 for that

.github/workflows/build.yml
```diff
@@ -0,0 +1,171 @@
...
+      uses: actions/checkout@v6
+
+    #####
+    # Dependency packages
+    #####
+
+    - name: Install dependecy packages
+      run: |
+        sudo apt update
+        sudo apt install -y openssl cmake zlib1g gcc
```
ohtanim: Do we really need those installation ? AFAIK, at least gcc and cmake has already been available in the runner VM image provided by Git.

.github/workflows/build.yml
```diff
@@ -0,0 +1,171 @@
...
+    - name: Install python dependencies
+      run: pip install -r requirements-dev.txt
+
+    #####
+    # Python lint check
+    #####
+
+    # TODO: Remove continue-on-error once all pylint warnings are fixed
+    - name: pylint
+      run: pylint ./python
```
ohtanim: This is incorrect sequence of steps. Before executing this step, you should run 
`maturin build --release`
and 
`pip install target/wheels/qrmi-*.whl`
to install the dependencies of /python code, such as qiskit_ibm_runtime etc..

.github/workflows/unittest-rust.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    - name: Unittest stubgen
+      run: cargo test --bin stubgen --release --features="pyo3"
+
+    - name: Unittest direct_access_client
+      working-directory: dependencies/direct_access_client
+      run: cargo test --release
+
+    - name: Unittest pascal_cloud_client
+      working-directory: dependencies/pascal_cloud_client
```
OkuyanBoga: "pasqal"

.github/workflows/build-c-examples.yml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    steps:
+
+    - name: Checkout
+      uses: actions/checkout@v6
+
+    # the qrmi.h is required to build the C examples
+    - name: Download libqrmi-artifacts
```
ohtanim: I'm not sure how we run this build-c-example workflow, but this workflow execution depends on the completion of build-qrmi workflow. In this case, I would recommend to define "job dependencies". 

Job dependencies cannot be defined across multiple workflow yaml, so I would suggest to move build-c-example workflow to build-libqrmi workflow.

What do you think ?

ohtanim: @cclaudio ^

cclaudio:
> Waiting for the response from @cclaudio .

Thank you, I just fixed that. For some reason, I only received today the comments that @ohtanim made two days ago. I am not sure if that's related to the github instability that happened a few days ago or if there is a problem in my email client

PR #48, opened 2026-02-18
Add Cargo.lock file to pin crate versions

ohtanim:
This PR adds a Cargo.lock file to pin the crate versions. The lock file was generated during the build process for testing with the latest crates available today. I believe this will benefit HPC use cases, which typically prefer not to upgrade versions too frequently. I also updated the .gitignore file because Cargo.lock had been listed as an ignored file.

If you already created lock file for clients/partners, please replace this one with yours. Thanks,

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -56,7 +56,7 @@ This section shows how to build QRMI for C and Python.
 ```shell-session
 . ~/.cargo/env
 cargo clean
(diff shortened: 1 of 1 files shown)


cclaudio:
LGTM.

We may need to udpate the documentation with:
```
cargo build --locked --release
```

Issue #34, opened 2025-08-06
Add resource_type()/resource_id() methods to QRMI QuantumResource


ohtanim:
### What problem are you trying to solve?

QuantumResource instances are created by QRMIService, but currently there is no way to determine which resource ID (backend name) is associated with them or what type of resource they are.

I think it is necessary to obtain this information from user code so that QRMI instances can be selected.

```
from qiskit_qrmi_primitives import QRMIService
from qiskit_qrmi_primitives.ibm import SamplerV2

service = QRMIService()
resources = service.resources()
for resource in resources {

}
```

### Describe the solution you’d like

Add the following 2 methods to QRMI Quantum Resource Trait

- async fn resource_id(&mut self) -> Result<String>
   Returns resource id of QuantumResource instance
- async fn resouce_type(&mut self) -> Result<ResourceType)>
   Returns the type of QauntumResource instance. For example, ResourceType::PasqalCloud

Equivalent methods should be available in C/Python bindings.

### Describe any alternatives you’ve considered

Or as alternative, QuantumResource::metadata() must contain those information with predefined hashmap keys.

awennersteen:
Has this been done?

ohtanim:
> Has this been done?

Not yet.

awennersteen:
I'm assigning myself then. It's been a while since I looked at all the "core" parts so I want to implement this to remind myself.


It will come after #32 and #33 have been merged

PR #56, opened 2026-02-25
Add resource_id() and resource_type() methods to QRMI

awennersteen:
## Description of Change

This PR introduces methods to determine which resource ID (backend name) or what type of resource a given `QuantumResource` is.

We add the following 2 methods to QRMI Quantum Resource Trait

```rust
    async fn resource_id(&mut self) -> Result<String>
    // Returns resource id of QuantumResource instance. For example, `"FRESNEL"`.
    async fn resouce_type(&mut self) -> Result<ResourceType)>
    // Returns the type of QauntumResource instance. For example, `ResourceType::PasqalCloud`.
```

I've implemented it for the ones that are already in the repo: `PasqalCloud`, `IBM DA` and `Runtime`. I've tested for `PasqalCloud`, and `IBM DA`, but I'm not able to test against `IBM Runtime`.

I modified all examples to include this print since it's in the QRMI spec, so the PR touches many files although the core change is small. (edited)

## Checklist ✅

- [ x ] Have you included a description of this change?
- [ x ] Have you updated the relevant documentation to reflect this change?
- [ x ] Have you made sure CI is passing before requesting a review?

## Ticket
Fixes https://github.com/qiskit-community/qrmi/issues/34

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,38 +27,36 @@ use async_trait::async_trait;
 /// Defines interfaces to quantum resources.
 #[async_trait]
 pub trait QuantumResource: Send + Sync {
-    /// Returns resource identifier of this quantum resource.
     async fn resource_id(&mut self) -> Result<String>;
-
-    /// Returns resource type of this quantum resource.
-    async fn resource_type(&mut self) -> Result<ResourceType>;
-
-    /// Returns true if device is accessible, otherwise false.
-    ///
-    /// # Arguments
-    ///
-    /// * `id`: Identifier of quantum device.
+    /// Returns resource identifier of this quantum resource.
     ///
     /// # Example
     ///
     /// ```no_run
     /// fn main() -> Result<(), Box<dyn std::error::Error>> {
-    ///     let mut qrmi = qrmi::QiskitRuntimeService::new("ibm_torino");
+    ///     let mut qrmi = IBMQiskitRuntimeService::new("ibm_torino")?;
+    ///     let resource_id = qrmi.resource_id().await?;
+    ///     println!("{resource_id}"); // prints "ibm_torino"
+    ///     Ok(())
+    /// }
+    /// ```
+    async fn resource_type(&mut self) -> Result<ResourceType>;
     ///
-    ///     let accessible = qrmi.is_accessible()?;
-    ///     if accessible == false {
-    ///         panic!("ibm_torino is not accessible.");
-    ///     }
+    /// Returns resource type of this quantum resource.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    ///     let mut qrmi = IBMQiskitRuntimeService::new("ibm_torino")?;
+    ///     let resource_type = qrmi.resource_type().await?;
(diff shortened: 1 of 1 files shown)


awennersteen:
I should probably bump qrmi version to 0.11.0 with this since we're changing QRMI spec

ohtanim:
> I should probably bump qrmi version to 0.11.0 with this since we're changing QRMI spec

@awennersteen I agreed. I would like to release 0.11.0 for this specification change.

BTW, are you going to include Pasqal's recent changes to 0.11.0 as well ?

examples/qiskit_primitives/pasqal/sampler.py
```diff
@@ -28,6 +28,11 @@
 if len(resources) == 0:
     raise ValueError("No quantum resource is available.")
 
+for res in resources:
```
ohtanim: Code formatting error was detect by "black sampler.py"

```
--- a/examples/qiskit_primitives/pasqal/sampler.py
+++ b/examples/qiskit_primitives/pasqal/sampler.py
@@ -29,9 +29,7 @@ if len(resources) == 0:
     raise ValueError("No quantum resource is available.")
 
 for res in resources:
-    print(
-        f"Available resource: id={res.resource_id()} type={str(res.resource_type())}"
-    )
+    print(f"Available resource: id={res.resource_id()} type={str(res.resource_type())}")
 
 # Randomly select QR
 qrmi = resources[random.randrange(len(resources))]
@@ -68,4 +66,4 @@ qc.append(gate, qc.qubits)
 
 sampler = QPPSamplerV2(qrmi=qrmi)
 res = sampler.run([qc])
-print(json.loads(res[0])['counter'])
+print(json.loads(res[0])["counter"])
```

src/lib.rs
```diff
@@ -20,13 +20,19 @@ pub mod models;
...
 use async_trait::async_trait;
 
 /// Defines interfaces to quantum resources.
 #[async_trait]
 pub trait QuantumResource: Send + Sync {
+    /// Returns resource identifier of this quantum resource.
+    async fn resource_id(&mut self) -> Result<String>;
+
+    /// Returns resource type of this quantum resource.
+    async fn resource_type(&mut self) -> Result<ResourceType>;
```
ohtanim: I would like to add # Example block to the comment like being done in other functions, but I'm ok to postpone this rustdoc update.

ohtanim:
> I will work on merging some of them right now, and i'll see if I put all of it in a new version, or keep this one as 0.11 and then bring in all our changes as 0.12 since they're local to us but quite large.
> 
> If you have a preference, let me know

Shall we go ahead and publish a version that includes the API spec changes?
That might also be preferable from a change‑management perspective. Let’s merge what we can into the main branch and release 0.11.0.

awennersteen:
Thanks @ohtanim 

Sorry about the black issues, turns out that I didn't run it with standard rules applied.

src/lib.rs
awennersteen: updated here

awennersteen:
> Shall we go ahead and publish a version that includes the API spec changes?
> That might also be preferable from a change‑management perspective. Let’s merge what we can into the main branch and release 0.11.0.

yes, agree

awennersteen:
I added @OkuyanBoga's suggestion as an issue in a  more general way through a developer guide.

PR #64, opened 2026-02-26
Upload libqrmi.so and qrmi.h artifacts for testing/debugging

cclaudio:
- Upload the libqrmi.so and qrmi.h built on pull-requests. This can help developers to further test/debug the artifacts in a local environment.
- Set the uploaded artifacts to be deleted after 5 days. AI suggests a retention-days between 1-7 days

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -48,7 +48,9 @@ jobs:
 
     - name: Upload QRMI wheels
       uses: actions/upload-artifact@v6
+      if: ${{ success() }}
       with:
         name: qrmi-wheels
         path: target/wheels/qrmi-*.whl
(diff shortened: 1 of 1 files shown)


ohtanim:
I’ll approve this for now, but I don’t understand which use cases in HPC software development this is supposed to help.In HPC, processors, compilers, and development environments vary from one HPC data center to another. In practice, they use RHEL 8/9 or CentOS. Compilers are also switched between various versions using environment modules or Spack. These are different from the ones bundled with the OS mentioned above and are typically optimized for the specific processors and architectures. JSC made a similar comment last year. This is also why I haven’t been creating and uploading artifacts.Even if we were to upload artifacts here, wouldn’t developers end up rebuilding them in their own environments anyway? I certainly would.

PR #67, opened 2026-02-27
Removal of openssl-sys dependencies

ohtanim:
### Issues:

Currently, libqrmi.so internally uses an HTTP client crate called reqwest. By default, its TLS library is built to dynamically link against the system OpenSSL library that is used during the build (via the openssl‑sys crate). The version of the OpenSSL library it links to varies depending on the operating system. For example, on RHEL 8 only libssl.so.1.1 is available, while on later versions of RHEL, libssl.so.3 is used. These two versions cannot coexist.
This leads to the following problem: when checking the list of shared library dependencies of a libqrmi.so built on RHEL 8, it is shown to depend on libssl.so.1.1. 

```
ldd libqrmi.so 
	libssl.so.1.1 => /lib64/libssl.so.1.1 (0x00007ffffe254000)
	libcrypto.so.1.1 => /lib64/libcrypto.so.1.1 (0x00007ffffdd69000)
        :
        :
```

If this binary is brought to RHEL 9, RHEL 10, AlmaLinux, or Rocky Linux—which are binary‑compatible with each other—it results in a runtime error because libssl.so.1.1 does not exist on those systems. This becomes a major issue when considering binary distribution of QRMI.

### Solution:
Instead of depending on the system OpenSSL library, use the `rustls-tls` crate provided as a Rust library and link it at build time. This eliminates the need to worry about the platform’s `libssl.so` version, allowing `libqrmi.so` to be deployed and executed on any supported platform.
Below is an example of a `Cargo.toml` configuration. It disables reqwest’s default `native-tls` feature and enables `rustls-tls` instead:

After this modification, libssl.so and libcrypt.so will not be listed as dependent shared library like below.

```
ldd /shared/qrmi/target/release/libqrmi.so 
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007ffffe2b0000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007ffffe08e000)
	libm.so.6 => /lib64/libm.so.6 (0x00007ffffdd0c000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007ffffdb08000)
	libc.so.6 => /lib64/libc.so.6 (0x00007ffffd732000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fffffdce000)
```

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -8,7 +8,6 @@
   * Rust compiler 1.91 or above [Link](https://www.rust-lang.org/tools/install)
   * A C compiler: for example, GCC(gcc) on Linux and Clang(clang-tools-extra) for Rust unknown targets/cross compilations. QRMI is compatible with a compiler conforming to the C11 standard.
   * make/cmake (make/cmake RPM for RHEL compatible OS)
-  * zlib (zlib-devel RPM for RHEL compatible OS)
   * Python 3.11, 3.12 or 3.13 (For Python API)
     * Libraries and header files needed for Python development(python3.1x-devel RPM for RHEL compatible OS)
       * /usr/include/python3.1x
@@ -16,7 +15,6 @@
 
 * Runtime requires the following tools:
   * gcc (libgcc RPM for RHEL compatible OS)
-  * zlib (zlib RPM for RHEL compatible OS)
   * Python 3.11, 3.12 or 3.13 (For Python API)
     * Libraries and header files needed for Python development(python3.1x-devel RPM for RHEL compatible OS)
 


ohtanim:
I've tested this for both IBM Direct Access and IBM Qiskit Runtime service. 

RH8 build works as expected without rebuilding libqrmi.so on:
- RHEL 9, 10
- Rocky 8, 9, 10
- AlmaLinux 8, 9 and 10

awennersteen (approved):
lgtm

I've also verified that  libqrmi built in RHEL-based works on ubuntu 24

PR #68, opened 2026-03-02
[DRAFT-only-for-discussion] QRMI packaging

### LINKED ISSUE #65: Create a CI workflow to help in the release process
A new CI workflow could properly test the code before releasing artifacts.

The tests we are running on pull-requests are supposed to be more specific and fast. We can definetly cover more tests before releasing artifacts, for example, we could run workflows against a larger test matrix we want to support (other ubuntu versions or distros, other python versions, etc) and also run tests that takes longer to run such as integration tests (when we have them)

What artifacts do we want to publish?
- This is TBD. A simple artifact could the source code tarball. We can expand that to libraries, header file and binaries.

cclaudio:
Fixes: #65

- Add rpm and deb files for packaging
- Add Makefile as a single source of truth for building, testing, checking and packaging
- Add worflows for building rpm and deb
- Add workflow for releasing artifacts

Some discussion topics:
- I think Makefile is a good option to automate build system, at least for now. Let me know if you have a different suggestion
- `make rpm` and `make deb` can be used by users and the CI. If we want to publish rpm and deb files, we should define what distros (and distro versions) has higher priority. That can save some space


packaging/rpm/qrmi.spec
```diff
@@ -0,0 +1,142 @@
...
+License:        Apache-2.0
+URL:            https://github.com/qiskit-community/qrmi
+Source0:        %{crate}-%{version}.tar.gz
+Source1:        %{crate}-%{version}-vendor.tar.gz
+
+BuildRequires:  gcc
+BuildRequires:  gcc-c++
+BuildRequires:  make
+BuildRequires:  cmake
+BuildRequires:  openssl-devel
```
ohtanim: openssl-devel is no longer needed.

ohtanim (changes requested):

ohtanim (changes requested):

cclaudio:
This PR is outdated:
- We already have a Makefile
- [Some concerns](https://qiskit.slack.com/archives/C08ETNQDZK4/p1772491525823179?thread_ts=1772463917.284289&cid=C08ETNQDZK4) related to distributing rpm/deb packages

I am closing it.

PR #84, opened 2026-03-16
Modify the pull-request workflow to run within ubi8 container

### LINKED ISSUE #82: CI: Modify the pull-request workflow to run on ubi8
Currently, our pull-request CI is building the code for Ubuntu 24.04. We should change that to build the code for ubi8, which is the same distro we are releasing artifacts for.

cclaudio:
Fixes #82

Not all the packages we need in the CI are installed in the UBI8 container image.

Changes in this PR:
- Install from distro the packages we need for the CI
- Install python3.x and python3.x-devel whenever we run cargo with `--features="pyo3"`
- Modify the `on-pull-request.yml` workflow to run within a ubi8 container

### DIFF
--- a/.github/workflows/build-libqrmi.yml
+++ b/.github/workflows/build-libqrmi.yml
@@ -2,7 +2,7 @@
 # This reusable workflow builds the libqrmi.so and the C examples
 #
 # Workflow inputs:
-# - None
+# - [required] container-image
 #
 # Workflow artifacts produced:
 # - [uploaded] libqrmi.so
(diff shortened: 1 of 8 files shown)


requirements-distro.txt
```diff
@@ -0,0 +1,3 @@
+gcc
```
ohtanim: Could you please add the following packages ? AFAIK, munge will be optionally used by Pasqal releases.

- munge
- munge-devel

I'm just wondering if we need to install:
- ca-certificates

ohtanim (approved):
I agree with postponing the discussion about munge. Approved.

Issue #81, opened 2026-03-13
Add makefile targets to cover all the steps found in the CI and the INSTALL.md


cclaudio:
Currently, the CI workflows are doing many operations that developers would also be interested on running standalone (e.g. building, testing, etc).

We should use the Makefile as single source of truth for operations like building, testing, etc. With that, both users and the CI could just run `make <target>`.

PR #87, opened 2026-03-21
Add makefile targets to cover all steps found in the CI and INSTALL.md

cclaudio:
Fixes: #81 
Depends-on: #84

The command below describes all targets implemented.
```
make help
```

Once this PR is merged we can update the CI workflows and the INSTALL.md

### Known issue

~~`make test-doc` is failing. The rust doctests need to be fixed for them to pass.~~   -------- Fixed in #88

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,7 @@ qrmi.h
 *target
 
 # python venvs
-*venv
+*venv*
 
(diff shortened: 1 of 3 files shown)


awennersteen (approved):
lgtm. I think this should just go in. It will go stale and we can just incrementally update

PR #96, opened 2026-04-01
[PASQAL] Rename pulser_backend -> pulser and better abide by Pulser Backend design, improve qpp integration. minor fix in  task runner

awennersteen:
## Description of Change

QRMI Pasqal updates: Pulser rename, task-runner/script fixes, QPP 0.2.0 alignment, and payload parsing fixes

 This PR bundles several related Pasqal integration changes that ideally should have been separate PRs.
 It includes:

 1. pulser_backend → pulser rename/alignment updates.
 2. Task-runner and script fixes in the relevant repos (qrmi, spank-plugins, and QPP-facing paths).
 3. Qiskit Pasqal Provider updates for qiskit-pasqal-provider==0.2.0 behavior.
 4. Payload parsing/serialization fixes for Pasqal target and result payloads.

Matching spank-plugin PR: https://github.com/qiskit-community/spank-plugins/pull/191

  ## Why This Is One PR

 This branch was used to get the full Pasqal workflow green end-to-end and multitasking over time. In hindsight, this should have been split into 3–4 PRs (rename, task-runner/scripts, QPP upgrade/fixes, serialization fixes).

  ## Changes By Group

  ### 1) Pulser naming and integration updates

  - Updated references from pulser_backend to pulser where needed.
  - Kept existing runtime behavior while restoring device/job-params validation when target specs are available.

  ### 2) Task runner and job script updates

  - Updated the relevant run/setup scripts in qrmi / spank-plugins paths so they point at the current examples and workflow entrypoints.
  - Verified task-runner path behavior as part of end-to-end checks.

  ### 3) QPP 0.2.0 compatibility updates

  - Aligned QRMI Pasqal primitive behavior with current qiskit-pasqal-provider expectations.
  - Fixed result conversion path so QRMI-returned Pasqal payloads produce expected PasqalResult behavior.

  ### 4) Serialization / payload parsing fixes

  - Added support for Pasqal payload inputs as:
      - JSON text
      - dict objects
      - Pasqal model-like objects (model_dump / dict)

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #


python/qrmi/tools/task_runner/main.py
```diff
@@ -117,7 +117,7 @@ def _find_qpu_type(self, qpu_name: str) -> ResourceType:
         for index, qpu_resource in enumerate(qpu_resources):
             if qpu_resource == qpu_name:
                 return self.RESOURCE_TYPE_MAP[qpu_types[index]]
-        return ValueError(f"{qpu_name} is not available")
+        raise ValueError(f"{qpu_name} is not available")
```
awennersteen, line 121: @ohtanim this should be a `raise ValueError`, right?

ohtanim: That's exactly right. Thanks for fixing.

ohtanim (approved):
LGTM

python/qrmi/primitives/pasqal/target.py
```diff
@@ -16,25 +16,61 @@
 
 # pylint: disable=invalid-name
 
+import json
```
awennersteen, line 21: in reality this whole target approach should be re-worked now that I understand better how it works. but I'd rather do this in a followup and for now give a slightly better user-experience with the qpp.

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,351 @@
...
+            logging.warning(
+                "The selected connection doesn't give access to the latest "
+                "device specs. Execution might fail if the sequence is "
+                "incompatible with the device."
+            )
+            return sequence
+
+    def _fetch_result(
+        self, batch_id: str, job_ids: list[str] | None
+    ) -> pulser.Sequence[Results]:
```
awennersteen: this is almost certainly not the right type

badtst:
```suggestion
    ) -> typing.Sequence[Results]:
```

examples/pulser/pasqal/requirements.txt
badtst, line 1: Since `v0.13.4` Pasqal-specific dependencies are optional : https://github.com/qiskit-community/qrmi/commit/fae941dd2b39b8a1126e3e01714c4d6da7a50db1#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

Could move those requirements to just `qrmi[pasqal]`

examples/pulser/pasqal/send_pasqal_job_local.py
badtst: Move from RemoteBackend to more user-facing backend `QPUBackend` also used in `pulser_qrmi.py`

examples/pulser/pasqal/requirements.txt
ohtanim: @badtst Correct, but we are having some issues around GHA release workflow, so we still need to "pip install qrmi" when installing from pypi.

@cclaudio has been investigating for the PR #128.

badtst: Noted thx

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,351 @@
...
+        self,
+        sequence: pulser.Sequence,
+        wait: bool = True,
+        open: bool = False,  # pylint: disable=redefined-builtin
+        batch_id: str | None = None,
+        **kwargs: typing.Any,
+    ) -> RemoteResults | list[typing.Any]:
+        """Submits the sequence for execution on a remote Pasqal backend.
+
+        For compatibility with older QRMI Pulser examples:
```
badtst: Are there older QRMI Pulser examples taht we must support that are not present in this repo ?

examples/pulser/pasqal/pulser_qrmi.py
badtst, line 1: pulser_qrmi.py is almost identical to send_pasqal_job_local.py should we merge them in a single script ?

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,351 @@
...
+            while True:
+                status = self._qrmi.task_status(job_id)
+                if status == TaskStatus.Completed:
+                    raw_results.append(self._qrmi.task_result(job_id).value)
+                    break
+                if status in (TaskStatus.Failed, TaskStatus.Cancelled):
+                    break
+                time.sleep(1)
+
+        return raw_results
```
badtst: We could just return a list of `[self.get_results(...)]` here

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,351 @@
...
+            )
+            payload = Payload.PasqalCloud(
+                sequence=seq_to_submit.to_abstract_repr(), job_runs=params["runs"]
+            )
+            task_id = self._qrmi.task_start(payload)
+            logger.info("task start: %s", task_id)
+            self._task_sequences[task_id] = seq_to_submit
+            new_job_ids.append(task_id)
+
+        new_batch_id = new_job_ids[0] if len(new_job_ids) == 1 else str(uuid.uuid4())
```
awennersteen: Check if I need to fix this - bad idea?

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,351 @@
...
+                raise
+            logging.warning(
+                "The selected connection doesn't give access to the latest "
+                "device specs. Execution might fail if the sequence is "
+                "incompatible with the device."
+            )
+            return sequence
+
+    def _fetch_result(
+        self, batch_id: str, job_ids: list[str] | None
```
awennersteen, line 132: ?
```suggestion
        self, batch_id: str | None, job_ids: list[str] | None
```

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,351 @@
...
+            return BatchStatus.PENDING
+        if any(status == JobStatus.CANCELED for status in statuses):
+            return BatchStatus.CANCELED
+        return BatchStatus.DONE
+
+    def _query_job_progress(
+        self, batch_id: str
+    ) -> typing.Mapping[str, tuple[JobStatus, Results | None]]:
+        progress: dict[str, tuple[JobStatus, Results | None]] = {}
+        for job_id in self._get_job_ids(batch_id):
```
awennersteen, line 184: this makes no sense to me looking at this again

badtst: This looks Ok to me, it seems to correctly implement the RemoteConnection _query_job_progress interface. Maybe we should add the docstrings of the RemoteConnection interfaces

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,351 @@
...
+                except (
+                    RemoteResultsError,
+                    json.JSONDecodeError,
+                    KeyError,
+                    TypeError,
+                    ValueError,
+                ) as err:
+                    logger.warning(
+                        "Failed to parse QRMI result for job %s: %s", job_id, err
+                    )
```
badtst, line 199:
```suggestion
                except (Exception):
                    logger.warning(
                        "Failed to parse QRMI result for job %s", job_id, exc_info=True,
                    )
```

python/qrmi/pulser/connection.py
awennersteen: no, not needed

examples/pulser/pasqal/pulser_qrmi.py
awennersteen: this might be outdated folder structure on my end.

python/qrmi/primitives/pasqal/sampler.py
```diff
@@ -10,80 +10,225 @@
...
+        parsed_payload = json.loads(payload)
+    elif isinstance(payload, dict):
+        parsed_payload = payload
+    else:
+        raise TypeError("Unsupported payload type. Expected JSON string or dict.")
+
+    if not isinstance(parsed_payload, dict):
+        raise TypeError("Invalid payload. Expected a JSON object.")
+
+    return parsed_payload
```
awennersteen, line 60: check this, we fixed this in warden

python/qrmi/primitives/pasqal/sampler.py
```diff
@@ -10,80 +10,225 @@
...
+        raise RuntimeError(f"Unsupported counter payload: {counter_payload!r}.")
+    counts = {
+        str(bitstring): int(count)
+        for bitstring, count in counter_payload.items()
+        if isinstance(count, (int, float))
+    }
+    if not counts:
+        raise RuntimeError(f"No valid counts found in payload: {payload!r}.")
+    return counts
+
```
awennersteen, line 76: probably same fix

python/qrmi/pulser/connection.py
badtst: Why should we add None here ?

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,345 @@
...
+        dev = pulser.abstract_repr.deserialize_device(json.dumps(target_payload))
+        return {dev.name: dev}
+
+    def update_sequence_device(self, sequence: pulser.Sequence) -> pulser.Sequence:
+        """Match sequence device to remote specs when available."""
+        try:
+            return super().update_sequence_device(sequence)
+        except RuntimeError as err:
+            if not _is_missing_device_specs_error(err):
+                raise
```
badtst: Is this what you meant @awennersteen ? 
```suggestion
                raise err
```

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,345 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _normalize_json_payload(payload: typing.Any) -> dict[str, typing.Any]:
+    """Return payload as a dictionary from JSON text or dict."""
+    if isinstance(payload, str):
+        normalized = json.loads(payload)
+    elif isinstance(payload, dict):
+        normalized = payload
```
badtst: From my tests with `mock-qpu-api` (with the latest fix to the mock resuilts) and `pasqos` emulator, running the `pulser_qrmi.py` example, this branch is never hit, which is the expected behavior

python/qrmi/pulser/connection.py
awennersteen: in the case of warden, there is no batch_id, right? 
And also from memory the interface is `str | None`.

badtst: https://github.com/pasqal-io/Pulser/blob/a4c7f1543848a060fa225a54adef5c7fe0419eec/pulser-core/pulser/backend/remote.py#L178 Interface is indeed `str`

badtst: The `batch_id` passed to this function is the one that you pass to the `RemoteResults` object init, which for now is `new_batch_id`

badtst: Indeed we do not use  `batch_id` for submitting to warden, but the connection uses it to store associated jobs.

badtst (changes requested):
QRMIConnection's behavior seems to work ok now. Still need a cleaning round .

I also think we need to check how we could integrate the recently merged `task_logs` interface within the examples and/or the connection

examples/pulser/pasqal/target.py
badtst, line 1: We should be able to get rid of this file and use the `fetch_available_devices` method from the qrmi connection instead

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,348 @@
...
+        )
+
+    def supports_open_batch(self) -> bool:
+        """Flag to confirm this class doesn't support open batch creation."""
+        return False
+
+    def fetch_available_devices(self) -> dict[str, pulser.devices.Device]:
+        target = self._qrmi.target()
+        target_payload = _normalize_target_payload(target)
+        dev = pulser.abstract_repr.deserialize_device(json.dumps(target_payload))
```
badtst: I don't suppose we're expecting to use a qrmi other then a Pasqal one, so maybe we can enforce a check that the QRMI is on Pasqal kind (local/cloud) in the __init__ and then get rid of any check that we have afterwards ?

awennersteen: >I don't suppose we're expecting to use a qrmi other then a Pasqal one
Agreee

> o maybe we can enforce a check that the QRMI is on Pasqal kind (local/cloud) in the init 
sounds like a great idea

>  get rid of any check that we have afterwards?
I don't quite follow but I totally agree with the idea of only doing the checks on init. I guess this is the normalization? it is probably no longer needed after you fixed that upstream issue

badtst: Done

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+        """Fetches the devices available through this connection."""
+
+        # Pasqal local/cloud implementation deserialize specs as strings already
+        dev_str = self._qrmi.target().value
+        try:
+            dev = deserialize_device(dev_str)
+        except DeserializeDeviceError as err:
+            logger.error(f"Failed to deserialize device: {dev_str}")
+            raise RuntimeError from err
+        return {dev.name: dev}
```
badtst: Check behavior in cloud context, do we return the specs corresponding to the qpu requested by user ?

MatthieuMoreau0: careful how we handle emulators since they don't have device specs

examples/pulser/pasqal/pulser_qrmi.py
MatthieuMoreau0: If possible let's have a single script which works both for cloud and local

README.md
```diff
@@ -128,8 +128,10 @@ Full example is available [here](./examples/qrmi/python/ibm_quantum_system/examp
 
 Python QRMI can be used to implement Qiskit primitives (Sampler and Estimator).
 See example of Qiskit primitives here [for IBM backends](./examples/qiskit_primitives/ibm/sampler.py) or [for Pasqal machines](./examples/qiskit_primitives/pasqal/sampler.py).
+The Pasqal example uses provider `SamplerV2` with QRMI backend integration and returns a
+Qiskit result object via `job.result()`.
 
-See [example](./examples/pulser_backend/pasqal/pulser_backend.py) of QRMI working with Pasqal Pulser.
+See [example](./examples/pulser/pasqal/pulser_qrmi.py) of QRMI working with Pasqal Pulser.
```
MatthieuMoreau0: docnit: This section is a bit confusing. We are talking about SamplerV2 but the example is only using Pulser

Should the link be `examples/qiskit_primitives/pasqal/sampler.py` instead?

python/qrmi/primitives/pasqal/target.py
```diff
@@ -16,25 +16,61 @@
...
     Returns:
         pulser.devices.Device: Pulser device
     """
-    target = qrmi.target()
-    return pulser.abstract_repr.deserialize_device(target)
+    try:
+        target = qrmi.target()
+    except RuntimeError as err:
+        if _is_missing_device_specs_error(err):
+            return DigitalAnalogDevice
```
MatthieuMoreau0: Let's log it here

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+
+_QRMI_TASK_STATUS_MAP: dict[TaskStatus, JobStatus] = {
+    TaskStatus.Queued: JobStatus.PENDING,
+    TaskStatus.Running: JobStatus.RUNNING,
+    TaskStatus.Completed: JobStatus.DONE,
+    TaskStatus.Failed: JobStatus.ERROR,
+    TaskStatus.Cancelled: JobStatus.CANCELED,
+}
+
+JOB_EXEUTION_POLLING_INTERVAL_S = 1
```
MatthieuMoreau0:
```suggestion
JOB_EXECUTION_POLLING_INTERVAL_S = 1
```

python/qrmi/pulser/connection.py
MatthieuMoreau0: Let's get rid of it then. We should be confident in our API contract instead of implementing safeguards client side

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+    def fetch_available_devices(self) -> dict[str, Device]:
+        """Fetches the devices available through this connection."""
+
+        # Pasqal local/cloud implementation deserialize specs as strings already
+        dev_str = self._qrmi.target().value
+        try:
+            dev = deserialize_device(dev_str)
+        except DeserializeDeviceError as err:
+            logger.error(f"Failed to deserialize device: {dev_str}")
+            raise RuntimeError from err
```
MatthieuMoreau0: Raise more appropriate error

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+        return {dev.name: dev}
+
+    def update_sequence_device(self, sequence: pulser.Sequence) -> pulser.Sequence:
```
MatthieuMoreau0: I don't think we need this if we properly handle the error in fetch_available_devices

Pulser already takes care of the logging

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+        # Check that Job Params are correctly defined
+        job_params: list[JobParams] = pulser.json.utils.make_json_compatible(
+            kwargs.get("job_params", [])
+        )
+        mimic_qpu: bool = kwargs.get("mimic_qpu", False)
+        if mimic_qpu:
+            # Replace the sequence's device by the QPU's
+            sequence = self.update_sequence_device(sequence)
+            # Check that the job params match with the max number of runs
+            pulser.QPUBackend.validate_job_params(job_params, sequence.device.max_runs)
```
MatthieuMoreau0: This is already done by the `RemoteBackend` interface, no need to implement it in the `Connection`

https://github.com/pasqal-io/Pulser/blob/develop/pulser-core/pulser/backend/remote.py#L364

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+
+        # In PasqalCloud, if batch_id is not empty, we can submit new jobs to a
+        # batch we just created. This is not implemented in QRMI.
+        if batch_id:
+            raise NotImplementedError(
+                "It is not possible to add jobs to a previously created batch "
+                "with QRMI."
+            )
```
MatthieuMoreau0: This is also related to the open batch feature. Let's group this with the error message above

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+                None,
+            )
+            if target_device is None:
+                raise ValueError(
+                    f"The Sequence's device {sequence.device.name} doesn't match the "
+                    "name of a device of any available QPU. Select your device among "
+                    "fetch_available_devices() and change your Sequence's device using "
+                    "its switch_device method."
+                )
+            pulser.QPUBackend.validate_job_params(job_params, target_device.max_runs)
```
MatthieuMoreau0, line 241: Unless I misunderstand, this is already done by the RemoteBackend in pulser so this is unnecessary

edit: I see we do something similar to this in pulser-myqlm; so maybe double check with Antoine Cornillot to make sure I am not missing something

python/qrmi/pulser/connection.py
MatthieuMoreau0: In pulser-myqlm this is how we generate the batch ID. I suggest we do the same for consistency

```
    @staticmethod
    def _batch_id_from_job_ids(job_ids: list[str]) -> str:
        """Generate the Batch ID from a list of Job IDs."""
        return "|".join(job_ids)
```

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+    def _wait_job_execution(self, job_id: str):
+        "Waits until job execution is complete"
+        while True:
+            status = self._qrmi.task_status(job_id)
+            if status in (
+                TaskStatus.Completed,
+                TaskStatus.Failed,
+                TaskStatus.Cancelled,
+            ):
+                break
```
MatthieuMoreau0: nit: return here to skip the extra sleep

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+            time.sleep(JOB_EXEUTION_POLLING_INTERVAL_S)
+
+    def _get_job_ids(self, batch_id: str) -> list[str]:
```
MatthieuMoreau0, line 286: If we use the concatenation suggested above for the batch ID, then we can try to split it and retrieve the job IDs using the QRMI

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+        return tuple(results)
+
+    def _get_batch_status(self, batch_id: str) -> BatchStatus:
+        """Gets the status of a batch from its ID."""
+        statuses = [
+            self._to_job_status(self._qrmi.task_status(job_id))
+            for job_id in self._get_job_ids(batch_id)
+        ]
+        if not statuses:
+            return BatchStatus.DONE
```
MatthieuMoreau0: It sounds like an unexpected case if no job status was retrieved, I would suggest raising an error here

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+        if any(status == JobStatus.ERROR for status in statuses):
+            return BatchStatus.ERROR
+        if any(status == JobStatus.RUNNING for status in statuses):
+            return BatchStatus.RUNNING
+        if any(status == JobStatus.PENDING for status in statuses):
+            return BatchStatus.PENDING
+        if any(status == JobStatus.CANCELED for status in statuses):
+            return BatchStatus.CANCELED
```
MatthieuMoreau0: I don't think we should mark the full batch as canceled as soon as the job is canceled; imo it should be running or done in that case.

python/qrmi/pulser/connection.py
```diff
@@ -0,0 +1,359 @@
...
+    def _get_batch_status(self, batch_id: str) -> BatchStatus:
+        """Gets the status of a batch from its ID."""
+        statuses = [
+            self._to_job_status(self._qrmi.task_status(job_id))
+            for job_id in self._get_job_ids(batch_id)
+        ]
+        if not statuses:
+            return BatchStatus.DONE
+        if any(status == JobStatus.ERROR for status in statuses):
+            return BatchStatus.ERROR
```
MatthieuMoreau0: In the cloud we only mark the batch as error if all its jobs failed, I suggest we do the same here. Otherwise it is running/done

python/qrmi/pulser/connection.py
awennersteen: Right now we have 1 job == 1 batch though. So in general we don't have great handlng of that. 
Hence I handle them as the same. 

I think you have the best judgement here though

README.md
awennersteen: Isn't this just switching form documenting Qiskit interface to Pulser interface?

python/qrmi/pulser/connection.py
MatthieuMoreau0: The RemoteBackend `run` method takes a list of JobParams as input, so there can be multiple jobs sent at once. From the RemoteBackend point of view these jobs are batched together (even if they are not behind the scene in the cloud/warden) so the status must reflect all the jobs statuses

examples/pulser/pasqal/pulser_qrmi.py
badtst: https://github.com/qiskit-community/spank-plugins/pull/191

python/qrmi/primitives/pasqal/target.py
badtst: This was used I suppose to return a `DigitalAnalogDevice` whenever we were targetting an emulator, now that we changed the `target()` behavior we need to device what to do here. Do we always return a DigitalAnalogDevice or real specs from the cloud that we got through the `target()`

MatthieuMoreau0 (approved):
LGTM

PR #101, opened 2026-04-09
[FEATURE|PASQAL] Update pasqal local 'is_accessible' method to get data from pasqal middleware

badtst:
## Description of Change

Instead of always returning "Ok(true)" for the `is_accessible` interface in Pasqal local qrmi implementation, we delegate the information to the pasqal local middleware "Warden" [at this endpoint](https://github.com/pasqal-io/warden/blob/d1ae4ea6e1e99a9dba293685c9c9fceaf973b285/warden/api/routes/accessible.py#L12). 

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/src/pasqal/local.rs
+++ b/src/pasqal/local.rs
@@ -68,14 +68,8 @@ impl QuantumResource for PasqalLocal {
     }
 
     async fn is_accessible(&mut self) -> Result<bool> {
-        match self
-            .api_client
-            .get_accessible()
-            .await
-        {
-            Ok(accessible) => {
-                Ok(accessible.is_accessible)
-            }
+        match self.api_client.get_accessible().await {
+            Ok(accessible) => Ok(accessible.is_accessible),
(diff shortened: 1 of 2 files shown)


src/pasqal/local.rs
```diff
@@ -68,8 +68,16 @@ impl QuantumResource for PasqalLocal {
...
+        match self
+            .api_client
+            .get_accessible()
+            .await
+        {
+            Ok(accessible) => {
+                Ok(accessible.is_accessible)
+            }
+            Err(err) => Err(err),
+        }
```
badtst: If we return Ok(false) we can't supply a reason ?

ohtanim (approved):
Looks good to me.

awennersteen (approved):
lgtm, thanks!

MatthieuMoreau0 (approved):

PR #109, opened 2026-04-14
CI release: Remove job that publish to test.pypi.org

### LINKED ISSUE #108: Make publish-testpypi workflow idempotent
### Problem

The publish-testpypi works only on the first run.
When re-running the workflow, it fails due to the policy of testpypi, which is desribed in https://test.pypi.org/help/#file-name-reuse.

testpypi returns 400 bad request if already used.

## Expected behavior

The action should be idempotent and safely re-runnable.
Re-running the workflow should skip this error and move forward to subsequent workflow steps.

cclaudio:
## Description of Change

Once a package version is published to either pypi.org or test.pypi.org,
it cannot be overwritten or deleted/reused. So, it does not make sense
to publish to test.pypi.org in a workflow also used to publish to
pypi.org. If the job fails while publishing to test.pypi.org, we  would
need to bump the qrmi version to run the workflow again, however,
ultimately those versions would never exist in the actual pypi.org.

This PR removes the job that publishes to test.pypi.org, but adds a job
to check if the version published to pypi.org can be installed and
downloaded correctly. That could help catching and addressing issues
early in the version just published.

## Checklist ✅

## Ticket
- [x] Fixes #108 
- [ ] Is Part of #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -91,7 +91,7 @@ jobs:
   publish-github:
     if: github.repository_owner == 'qiskit-community'
     name: Publish github release
-    needs: [valid-qrmi-version, create-draft-release, check-testpypi-release]
+    needs: [valid-qrmi-version, create-draft-release]
     permissions:
       contents: write # This is required to publish the draft release
     runs-on: ubuntu-24.04
@@ -152,13 +152,13 @@ jobs:
         if-no-files-found: error
         retention-days: 1
 
-  publish-testpypi:
+  publish-pypi:
     if: github.repository_owner == 'qiskit-community'
-    needs: [build-and-test-wheels, wheels-sdist]
-    runs-on: ubuntu-24.04
+    needs: [build-and-test-wheels, wheels-sdist, publish-github]
+    runs-on: ubuntu-latest
     environment:
-      name: testpypi
-      url: https://test.pypi.org/project/qrmi/
+      name: pypi
+      url: https://pypi.org/project/qrmi
     permissions:
       id-token: write  # IMPORTANT: this permission is mandatory for trusted publishing
     steps:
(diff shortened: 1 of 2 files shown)


ohtanim:
My understandings are:
- publish-testpypi and check-testpypi-release were removed. 
- check-pypi-post-release was added. It covers what check-testpypi-release was done.
- all workflow steps can be executed if valid-qrmi-version is succeeded. This is probably designed this way to allow restarting from any step.

Curret DAG
image


New DAG
image

ohtanim (approved):
LGTM, but I have a question.

.github/workflows/release.yml
```diff
@@ -170,13 +173,16 @@ jobs:
...
 
-  check-testpypi-release:
+  # Once a package version is released to either pypi.org or test.pypi.org,
+  # it cannot be overwritten or deleted/reused. This job checks if the
+  # version released to pypi can be installed and downloaded correctly.
+  # That could help to catch and address issues early.
+  check-pypi-post-release:
     if: github.repository_owner == 'qiskit-community'
-    needs: [publish-testpypi, valid-qrmi-version]
+    needs: [publish-pypi, valid-qrmi-version]
```
ohtanim, line 189: I'm curious about why valid-qrmi-version is listed as a dependency for this step. What use cases are you expecting this specification to support?

cclaudio: Right, I think I can further improve that. It looks redundant.

Today I realized that after the `valid-qrmi-version` job has failed, some other jobs continued to run. However, I think only the leaves in the dependency tree need to depend on the `valid-qrmi-version`. I will fix that tomorrow. The `publish-pypi` may not need to depend on `valid-qrmi-version` as well. Thanks

cclaudio: Updated the branch:
- removed the `valid-qrmi-version` dependency from the `publish-pypi` job
- added some comments to `valid-qrmi-version`

@ohtanim the job `check-pypi-post-release` depends on `valid-qrmi-version` to provide the qrmi version that pip should try to install and download. So I could not remove the `valid-qrmi-version` dependency from it.

cclaudio: > all workflow steps can be executed if valid-qrmi-version is succeeded. This is probably designed this way to allow restarting from any step.
 
One goal of the `valid-qrmi-version` job is to query github and pypi to check if the current qrmi version has already been released. If so, there is no reason to start the subsequent jobs because the publish will fail anyway (version already exists). That's why all other jobs depend directly or indirectly on `valid-qrmi-version`.

PR #121, opened 2026-04-23
Enable release wheel builds on aarch64 and ppc64le

### LINKED ISSUE #116: release.yml: Build and publish wheels for ppc64le and arm64

We don't have pre-built wheels for ppc64le and arm64 yet.

Extend the release.yml workflow to build and publish wheels for ppc64le and arm64
Ideally, this should be done only after we have a CI for testing them (#77 - weekly tests)

_No response_

_No response_

mtreinish:
## Description of Change

This commit adds two new jobs to the wheels release job that will build precompiled packages of linux for aarch64 and ppc64le platforms. Both CPU architectures have some presence in HPC systems and providing packages for those clusters will provide benefits to users leveraging quantum computing from those architectures. The arm builds run on github's arm hosted runners and are able to run tests and and the whole cibuildwheel workflow. For ppc64le however github doesn't have a hosted runner solution available for it. To build those wheels we run the manylinux docker image under qemu emulation of the platform. This is exceedingly slow in practice so we limit that wheel job to just compilation and building the wheel. This will skip running the tests which is unavoidable as it could take longer than the job timeout (6 hours) to run the full test suite under emulation.

## Checklist ✅

## Ticket
- [X] Fixes #116
- [ ] Is Part of #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -114,7 +114,13 @@ jobs:
   build-and-test-wheels:
     if: github.repository_owner == 'qiskit-community'
     needs: [valid-qrmi-version]
-    runs-on: ubuntu-24.04
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-24.04
+          - ubuntu-24.04-arm
     steps:
     - uses: actions/checkout@v6
     - uses: actions/setup-python@v6
@@ -132,7 +138,37 @@ jobs:
       uses: actions/upload-artifact@v7
       if: ${{ success() }}
       with:
-        name: release-wheels-x86_64
+        name: release-wheels-${{ matrix.os }}
+        path: ./wheelhouse/*.whl
+        if-no-files-found: error
+        retention-days: 1
+  crossbuild-ppc64_wheels:
+    if: github.repository_owner == 'qiskit-community'
+    needs: [valid-qrmi-version]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-python@v6
+      with:
+        python-version: 3.12
+    - uses: docker/setup-qemu-action@v4
+      with:
+        platforms: all
+    # The manylinux image comes with various cpython versions installed.
+    # We use the cpython 3.12 to build the wheels, as defined in the CIBW_BUILD.
+    # The generated wheels is abi3 compatible, so it is compatible with other python versions.
+    - name: Build QRMI wheels
+      uses: pypa/cibuildwheel@v3.4.1
+      env:
+        CIBW_PLATFORM: linux
+        CIBW_BUILD: cp312-manylinux_ppc64le
+        CIBW_ARCHS_LINUX: ppc64le
+        CIBW_TEST_SKIP: "cp*"
+    - name: Upload QRMI wheels
+      uses: actions/upload-artifact@v7
+      if: ${{ success() }}
+      with:
+        name: release-wheels-ppc64le
         path: ./wheelhouse/*.whl
         if-no-files-found: error
         retention-days: 1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -121,54 +121,19 @@ jobs:
         os:
           - ubuntu-24.04
           - ubuntu-24.04-arm
+          - ubuntu-24.04-ppc64le
     steps:
     - uses: actions/checkout@v6
     - uses: actions/setup-python@v6
       with:
         python-version: 3.12
-    # The manylinux image comes with various cpython versions installed.
-    # We use the cpython 3.12 to build the wheels, as defined in the CIBW_BUILD.
-    # The generated wheels is abi3 compatible, so it is compatible with other python versions.
     - name: Build QRMI wheels
       uses: pypa/cibuildwheel@v3.4.1
-      env:
-        CIBW_PLATFORM: linux
-        CIBW_BUILD: cp312-manylinux_x86_64
     - name: Upload QRMI wheels
       uses: actions/upload-artifact@v7
       if: ${{ success() }}
       with:
         name: release-wheels-${{ matrix.os }}
-        path: ./wheelhouse/*.whl
-        if-no-files-found: error
-        retention-days: 1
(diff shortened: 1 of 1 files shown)


ohtanim:
@cclaudio Could you please take a look at this PR?

.github/workflows/release.yml
```diff
@@ -132,7 +138,37 @@ jobs:
       uses: actions/upload-artifact@v7
       if: ${{ success() }}
       with:
-        name: release-wheels-x86_64
+        name: release-wheels-${{ matrix.os }}
+        path: ./wheelhouse/*.whl
+        if-no-files-found: error
+        retention-days: 1
+  crossbuild-ppc64_wheels:
```
cclaudio: The `ubuntu-24.04-ppc64le` self-hosted runner is available in the qrmi repository. We can probably use it to build and test the wheel for ppc64le, no need to cross build it.

.github/workflows/release.yml
```diff
@@ -132,7 +138,37 @@ jobs:
...
+      with:
+        platforms: all
+    # The manylinux image comes with various cpython versions installed.
+    # We use the cpython 3.12 to build the wheels, as defined in the CIBW_BUILD.
+    # The generated wheels is abi3 compatible, so it is compatible with other python versions.
+    - name: Build QRMI wheels
+      uses: pypa/cibuildwheel@v3.4.1
+      env:
+        CIBW_PLATFORM: linux
+        CIBW_BUILD: cp312-manylinux_ppc64le
```
cclaudio: I think we will always build and test for python 3.11, 3.12 and 3.13. Does it make sense to update the CIBW_BUILD here or put it in the pyproject.toml like below?

At the end of the day, the cibuildwheel will build a wheel only for 3.11 because it is abi3, but at least we would have it tested in all three python versions. For now, we don't have many python test cases, so I think it should be fine to run them in the release workflow

I am happy to do that in a separate PR if you want, but I'd like to check if you have an opinion on that. 

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 6d83826f369c..230291fb63ff 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ task_runner = "qrmi.tools.task_runner.main:run"
 
 [tool.cibuildwheel]
 manylinux-x86_64-image = "manylinux_2_28"
+build = "cp311-manylinux* cp312-manylinux* cp313-manylinux*"
 
 [tool.cibuildwheel.linux]
 before-all = 'curl -sSf https://sh.rustup.rs | sh -s -- -y'
```

.github/workflows/release.yml
```diff
@@ -132,7 +138,37 @@ jobs:
...
+  crossbuild-ppc64_wheels:
+    if: github.repository_owner == 'qiskit-community'
+    needs: [valid-qrmi-version]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-python@v6
+      with:
+        python-version: 3.12
+    - uses: docker/setup-qemu-action@v4
```
cclaudio: I am wondering if this is required only for cross building

.github/workflows/release.yml
mtreinish: I only did it this way because it's what was there for the existing x86_64 job and I assumed there was a reason for it. This argument is telling cibuildwheel to only build the wheel for 3.12 and then test it on all supported platforms. In general if you remove this argument it will build with python 3.11 (based on the required python version) and test against all compatible python versions. I very rarely see it specified this way, especially for an abi3 project as it's generally better to just use the oldest version for building as you can assert the abi3 compatibility level more easily that way.

.github/workflows/release.yml
mtreinish: This is only needed for cross-building, it enables qemu via docker and then the options are to install all the architectures supported for qemu. This is what cibuildwheel leverages internally to do the cross building. If we have native ppc64le hosts then we don't need this

.github/workflows/release.yml
mtreinish: I didn't realize you had signed up for IBM/actionspz on this repository. If the ppc64le runners are available then we can just add the runner to the matrix on the above job and it should all work fine without needing this extra job.

.github/workflows/release.yml
cclaudio: Got it, thank you.

So I think we can just remove the `CIBW_BUILD` parameter and let cibuildwheel build for python 3.11.
Since we don't need to cross-compile, we don't need a new job only for ppc64le. We could just add the ppc64le runner to the `build-and-test-wheels` job

mtreinish: Done in: https://github.com/qiskit-community/qrmi/pull/121/commits/250cf57fd8de24ed81810cb5825aa2c63f701ae5

.github/workflows/release.yml
mtreinish: Done in: https://github.com/qiskit-community/qrmi/pull/121/commits/250cf57fd8de24ed81810cb5825aa2c63f701ae5

cclaudio (approved):
LGTM, thank you.

ohtanim (approved):

ohtanim:
GHA release workflows all failed. It appears that this PR changes may be causing this degradation.

https://github.com/qiskit-community/qrmi/actions/runs/25409034406/job/74526369965
https://github.com/qiskit-community/qrmi/actions/runs/25409034406/job/74526369959
https://github.com/qiskit-community/qrmi/actions/runs/25409034406/job/74526369967

ohtanim:
Installing qiskit-2.4.1 was failed for x86_64/aarch64 platforms due to the following error:

```
ValueError: `py_limited_api='cp310'` not supported.
`Py_LIMITED_API` is currently incompatible with `Py_GIL_DISABLED`.
```

This error occurred during cp314t(=Free-Threaded Python) build, not occurred during cp314 build.
Qiskit 2.4.1 attempts to build using `py_limited_api='cp310'` (Stable ABI), but `Py_LIMITED_API` is currently not supported in Free-Threaded Python ([CPython issue #111506](https://github.com/python/cpython/issues/111506)).

We've not yet supported python 3.14, so I think we should skip cp314/cp314t build for now.

---
For ppc64le, python 3.12 seems to not available. Remove ppc64le from matrix might be tentative solutions for now.

Issue #117, opened 2026-04-21
Build libqrmi with munge support


cclaudio:
### What problem are you trying to solve?

Currently, the libqrmi we are building and publishing does not support pasqal_local

### Describe the solution you’d like

Add munge as default feature in Cargo.toml. This will also require installing `munge-libs` in the build environment, which will be easy to do after we replace the ubi8 container image by manylinux_2_28 (it has munge-libs available in the repository).

cclaudio:
I am replacing the ubi8 by manylinux_2_28 as part of #77

MatthieuMoreau0:
Hello! In the description, you mention adding munge as a default feature. We intentionally set munge as an optional dependency because from our understanding (to confirm?) it is not a required dependency in HPC cluster so it may not be available. Some HPC center without munge may still be interested in the QRMI for IBM direct access or Pasqal Cloud and making it a default who prevent that.

Have you considered building different version of the libqrmi with and without munge support? Thus HPC admins may pick the relevant one if they need to support pasqal-local.

On another note, what about the Pypi package? Do we want to build wheels with munge support as well?

Thank you :)

cclaudio:
Those are great questions. We started discussing about that in a different issue, which is outdated at this point.

> Hello! In the description, you mention adding munge as a default feature. We intentionally set munge as an optional dependency because from our understanding (to confirm?) it is not a required dependency in HPC cluster so it may not be available. Some HPC center without munge may still be interested in the QRMI for IBM direct access or Pasqal Cloud and making it a default who prevent that.

Build may not be an issue, we could assume that podman/docker is available in the system. Would that be reasonable? We just would need to update the Makefile

```
$ podman run --rm -it quay.io/pypa/manylinux_2_28_x86_64 dnf install -y munge-devel
AlmaLinux 8 - BaseOS                                                                                                                                                                                    12 MB/s |  63 MB     00:05    
AlmaLinux 8 - AppStream                                                                                                                                                                                 10 MB/s |  27 MB     00:02    
AlmaLinux 8 - Extras                                                                                                                                                                                    22 kB/s |  14 kB     00:00    
AlmaLinux 8 - PowerTools                                                                                                                                                                               4.1 MB/s | 5.8 MB     00:01    
Extra Packages for Enterprise Linux 8 - x86_64                                                                                                                                                         7.4 MB/s |  14 MB     00:01    
Dependencies resolved.
=======================================================================================================================================================================================================================================
[...]

PR #122, opened 2026-04-23
Make vendor specific dependencies optional

mtreinish:
## Description of Change

Most QRMI installations do not require all the vendor plugins at once. Typically an installation will only have access to a quantum system from one or two vendors. Even if they have more it won't be all of the QRMI plugins at once. Requiring that a user that has a pasqal system install all of IBM's dependencies or vice versa complicates the install by making the Python dependency for QRMI much larger than what is needed for an installation. This commit splits out the vendor specific dependencies as optional dependency groups on the python package. So a user with an IBM quantum system installs `qrmi[ibm]` a user with IBM and Pasqal can install `qrmi[ibm,pasqal]`. When other vendors are added in the future they can be handled in this way. There is also an `all` optional added so if the user pip installs `qrmi[all]` it will install all the vendor dependencies.

One option that this commit does not implement is add a more graceful exception type around the vendor plugin code (you can look at what Qiskit does with it's `MissingOptionalLibrary` class and the optionals module with decorators as a model) to give a better UX when a vendor plugin is used and the dependencies are not installed. For right now it will just be an ImportError when the vendor specific module is used and one of the now optional libraries can't be found.

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -128,6 +128,7 @@ jobs:
       env:
         CIBW_PLATFORM: linux
         CIBW_BUILD: cp312-manylinux_x86_64
+        CIBW_TEST_EXTRAS: all
     - name: Upload QRMI wheels
       uses: actions/upload-artifact@v7
       if: ${{ success() }}
--- a/.github/workflows/unittest-python.yml
+++ b/.github/workflows/unittest-python.yml
@@ -56,7 +56,7 @@ jobs:
       run: pip install -r requirements-dev.txt
 
     - name: Install QRMI wheels
-      run: pip install 'qrmi-*.whl[all]'
+      run: pip install "$(ls qrmi-*.whl)[all]"
 
     - name: Unittest with pytest
       run: pytest -v ./python
--- a/.github/workflows/unittest-python.yml
+++ b/.github/workflows/unittest-python.yml
@@ -56,7 +56,7 @@ jobs:
       run: pip install -r requirements-dev.txt
 
     - name: Install QRMI wheels
-      run: pip install qrmi-*.whl
+      run: pip install 'qrmi-*.whl[all]'
 
     - name: Unittest with pytest
       run: pytest -v ./python
--- a/.github/workflows/lint-fmt-python.yml
+++ b/.github/workflows/lint-fmt-python.yml
@@ -59,7 +59,7 @@ jobs:
       run: pip install -r requirements-dev.txt
 
     - name: Install QRMI wheels
-      run: pip install qrmi-*.whl
+      run: pip install "$(ls qrmi-*.whl)[all]"
 
     - name: Format check with black
       run: black --check ./python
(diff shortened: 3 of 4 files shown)


awennersteen (approved):
lgtm. 

cc @machjn for doing this in their MR

ohtanim:
Added @cclaudio as reviewer since GHA workflow has been updated by this PR.

.github/workflows/release.yml
```diff
@@ -128,6 +128,7 @@ jobs:
       env:
         CIBW_PLATFORM: linux
         CIBW_BUILD: cp312-manylinux_x86_64
+        CIBW_TEST_EXTRAS: all
```
cclaudio, line 131: This is a really nice idea, thanks!

I know my question is somewhat stretch (or out of context), but in the issue https://github.com/qiskit-community/qrmi/issues/117 we discussed about the munge package dependency for `pasqal-local`. The munge package may not be available in all distros, but I was wondering if the `optional-dependencies` in pyproject.toml could help somehow to give a better UX. Would you have any ideas?

cclaudio (approved):
LGTM.
I made only one comment, but it may not be related to this PR

ohtanim:
@cclaudio I think this would require changes to your workflow. Are you planning to handle this in a separate PR? At the moment, two workflows(lint-fmt-python/unittest-python) are failing.

.github/workflows/unittest-python.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       run: pip install -r requirements-dev.txt
 
     - name: Install QRMI wheels
-      run: pip install qrmi-*.whl
+      run: pip install 'qrmi-*.whl[all]'
```
ohtanim: I think 
`pip install "$(ls qrmi-*.whl)[all]"`
might be correct.

@cclaudio Please double check.

ohtanim: Updated unittest-python and lint-fmt-python to fix the GHA errors.

ohtanim: I will update README in separate PR later.

.github/workflows/release.yml
mtreinish: Yeah you could specify it as an optional dependency group so that users opt-in to installing the dependency. The other option that python package metadata allows is to specify environment markers on a dependency: https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers . This will let you specify in the dependency list something like install package foo if the installing platform is not windows. There are limits to the platform specification with environment markers but if the constraints on munge's platform support are expressible in them (I've used it in the past to limit a dependency for specific python versions, OS, and cpu architecture) then that is also a path to make it optional. The environment markers are compatible with optional dependencies too. So you can have a dependency in an optional extra that also has environment markers.

PR #126, opened 2026-04-30
Simplify Rust object re-exports in Python package

mtreinish:
## Description of Change

This commit simplifies the mechanism by which the Rust defined data structures are exposed to the root of the Python package. Previously the use of importlib and manually inserting values into globals is making a lot of underlying assumptions about the structure of the python environment to work and was also making something very simple overly complicated. The simplest way to handle the re-export is to just import them into the qrmi python namespace, just doing that and they'll be public as part of the Python qrmi package. This makes that change to correct this.

One thing I would typically recommend to pair with this is adding an `__all__` dunder attribute to the qrmi package to make it explicit which objects are publicly available in the package instead of relying on what's been imported or defined in a module. But I opted not to make this change in this PR because it requires deciding more explicitly what should be public and that is better handled in isolation.

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -33,7 +33,7 @@ unsafe-load-any-extension=no
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code
-extension-pkg-whitelist=
+extension-pkg-whitelist=qrmi._core
 
 
 [MESSAGES CONTROL]
--- a/python/qrmi/__init__.py
+++ b/python/qrmi/__init__.py
@@ -14,8 +14,4 @@
 
 """QRMI python package"""
 
-from importlib import import_module as _im
-
-_core = _im(__name__ + "._core")  # -> qrmi._core  (Rust)
-
-globals().update({k: v for k, v in _core.__dict__.items() if not k.startswith("_")})
+from ._core import *
--- a/python/qrmi/__init__.py
+++ b/python/qrmi/__init__.py
@@ -14,4 +14,4 @@
 
 """QRMI python package"""
 
-from ._core import *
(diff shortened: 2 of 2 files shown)


ohtanim:
@mtreinish GHA linter workflow reported the following error. 

```
Run pylint ./python
************* Module /__w/qrmi/qrmi/.pylintrc
.pylintrc:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'too-many-positional-arguments' (unknown-option-value)
************* Module qrmi
python/qrmi/__init__.py:17:0: E0401: Unable to import 'qrmi._core' (import-error)

```
Do you think ignoring import error is reasonable to change ? 

`from ._core import *  # pylint: disable=import-error`

ohtanim:
> @mtreinish GHA linter workflow reported the following error.
> 
> ```
> Run pylint ./python
> ************* Module /__w/qrmi/qrmi/.pylintrc
> .pylintrc:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'too-many-positional-arguments' (unknown-option-value)
> ************* Module qrmi
> python/qrmi/__init__.py:17:0: E0401: Unable to import 'qrmi._core' (import-error)
> ```
> 
> Do you think ignoring import error is reasonable to change ?
> 
> `from ._core import * # pylint: disable=import-error`

Looking at pylint’s issue tracker, it seems that many false positives like this one have already been reported, especially around C extension modules and type stubs. For now, we’ll ignore pylint errors in GHA and move forward. We’re also considering switching to a linter based on ruff in the near future.

ohtanim (approved):
Thanks for fixing. Approved.

PR #128, opened 2026-05-06
Fix build-and-test-wheels workflow step errors introduced by #121

ohtanim:
## Description of Change

Recent #121 causes workflow execution errors:
https://github.com/qiskit-community/qrmi/actions/runs/25409034406

This PR addresses tentative solutions for now:
- Skip build/test with free-threaded python (cp314t) to avoid an installation error of qiskit-2.4.1
- Disable ppc64le platform (python 3.12 is not available on our ppc64le runner)

@cclaudio is expected to review this PR.

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -121,7 +121,7 @@ jobs:
         os:
           - ubuntu-24.04
           - ubuntu-24.04-arm
-          - ubuntu-24.04-ppc64le
+        #  - ubuntu-24.04-ppc64le
     steps:
     - uses: actions/checkout@v6
     - uses: actions/setup-python@v6
@@ -129,8 +129,10 @@ jobs:
         python-version: 3.12
     - name: Build QRMI wheels
       uses: pypa/cibuildwheel@v3.4.1
+      # Skip free-threaded python for qiskit installation
       env:
         CIBW_TEST_EXTRAS: all
+        CIBW_SKIP: "cp*t-*"
     - name: Upload QRMI wheels
       uses: actions/upload-artifact@v7
       if: ${{ success() }}


.github/workflows/release.yml
```diff
@@ -121,16 +121,18 @@ jobs:
         os:
           - ubuntu-24.04
           - ubuntu-24.04-arm
-          - ubuntu-24.04-ppc64le
+        #  - ubuntu-24.04-ppc64le
```
mtreinish, line 124: If the ppc64le runner isn't compatible with using cibuildwheel right now. We can switch back to what I originally had in #121 and cross build the ppc64le wheels under qemu emulation. In practice the motivation I had for #121 was for ppc64le wheels so I'd like to make sure we get those built even if we don't test them.

Although if the issue is that py 3.12 is missing then we can change the version on L129 to be any version of python that works with cibuildwheel, which for cibuildwheel 3.4.1 is Python >= 3.11. All that we need to install python for in this job is to run cibuildwheel. Cibuildwheel itself will take care of installing the python for building and testing wheels. So if the ppc64le runner has any version of python installed >=3.11 we can use that and it should work fine.

mtreinish (commented):
Do we have a mechanism to manually trigger a test run of cibuildwheel? In my experience these are tricky to get right in the abstract and you really need to run the build on any configuration change to ensure you have things configured correctly.

ohtanim:
> Do we have a mechanism to manually trigger a test run of cibuildwheel? In my experience these are tricky to get right in the abstract and you really need to run the build on any configuration change to ensure you have things configured correctly.

As far as I know, currently no. And I totally agree with preparing this mechanism. Before kicking the release process, I would like to test build wheels workflow manually.

@cclaudio This might be the topic what we discussed today in DM.

.github/workflows/release.yml
cclaudio: > If the ppc64le runner isn't compatible with using cibuildwheel right now.

We don't have that yet, although it would be very useful.

> We can switch back to what I originally had in #121 and cross build the ppc64le wheels under qemu emulation. In practice the motivation I had for #121 was for ppc64le wheels so I'd like to make sure we get those built even if we don't test them.

Thanks for your feedback. I am further investigating the issue now and I will try to submit a fix by the end of the week.

I'm also making some changes to the other GHAs for us to be able to trigger them manually for testing before triggering the release workflow.

cclaudio: Just sumitted #129 for that

ohtanim:
Seems that this PR is no longer needed. Closing.

PR #132, opened 2026-05-25
Add docs test/example to qrmi::models::Config

mcbennet:
## Description of Change

This PR adds a docs example/test to the `Config` struct in `src/models/config.rs`. I added this to my local branch for illustrative purposes initially, though it seems worthwhile to add to the codebase to provide a compiler-verified example demonstrating how a user should instantiate the config file and access resources

## Checklist ✅

## Tests

- [x] Verified the example compiles via `cargo test --doc`.
- [x] Verified HTML renders correctly (locally) via `cargo doc`.
- [x] Change was formatted with `cargo fmt`.

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/src/models/config.rs
+++ b/src/models/config.rs
@@ -101,6 +101,22 @@ pub struct ResourceDefs {
 }
 
 /// QRMI configuration file
+///
+/// # Example
+///
+/// ```no_run
+/// use qrmi::models::Config;
+/// 
+/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+/// let config = Config::load("./qrmi_example_config.json")?;
+///
+/// if let Some(resource) = config.resource_map.get("ibm_osaka") {
+///     println!("Found resource: {}", resource.name);
+///     println!("Type: {:?}", resource.r#type);
+/// }
+/// # Ok(())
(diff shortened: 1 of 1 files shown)


src/models/config.rs
```diff
@@ -101,6 +101,22 @@ pub struct ResourceDefs {
...
+///
+/// ```no_run
+/// use qrmi::models::Config;
+///
+/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+/// let config = Config::load("./qrmi_example_config.json")?;
+///
+/// if let Some(resource) = config.resource_map.get("ibm_osaka") {
+///     println!("Found resource: {}", resource.name);
+///     println!("Type: {:?}", resource.r#type);
```
ohtanim, line 115: I would suggest to add:

```
for (key, value) in &resource.environment {
    println!("Environment variable: {}={}", key, value);
}
```

ohtanim (approved):
lgtm

src/models/config.rs
mcbennet: Thank you for the suggestion, @ohtanim.  Added

PR #141, opened 2026-06-09
[PASQAL] Update Pasqal Cloud `is_accessible` behavior to call public endpoint

badtst:
## Description of Change

We update the `is_accessible` call for pasqal-coud implementation to use a public API endpoint with no auth needed.
This removes the need for the Spank plugin to have any authentication for pasqal-cloud resources so that we can setup pasqal-cloud access in the user-env only.

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/dependencies/pasqal_cloud_client/src/client.rs
+++ b/dependencies/pasqal_cloud_client/src/client.rs
@@ -191,12 +191,15 @@ impl Client {
         Ok(())
     }
 
+    // Pasqal Cloud gets devices during `is_accessible` call.
+    // We we the public api endpoint with no auth token required
+    // so that the Spank plugin does not need any authentication setup.
     pub async fn get_device(&mut self, device_type: DeviceType) -> Result<GetDeviceResponseData> {
         let url = format!(
             "{}/core-fast/api/v1/devices?device_type={}",
             self.base_url, device_type,
         );
-        let resp: Response<Vec<GetDeviceResponseData>> = self.get(&url).await?;
+        let resp: Response<Vec<GetDeviceResponseData>> = self.get_raw(&url).await?;
 
         resp.data
             .into_iter()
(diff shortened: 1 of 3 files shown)


src/pasqal/cloud.rs
```diff
@@ -251,7 +251,7 @@ impl PasqalCloud {
         } else if let Some(token) = auth_token {
             builder.with_token(token);
         } else {
-            error!(
+            debug!(
```
badtst, line 254: Since the auth parameters are not needed for the Spank usage of the pasqal-cloud resource, this log should not be forwarded to the user. 

If the user fails to setup credentials, he gets another error log in `ensure_authenticated` further down the line

awennersteen: This only affects for the pasqal-cloud resource though, right?

I'm not sure I follow

badtst: Yes pasqal-cloud sorry, typo

awennersteen: ah gotcha, now I see. so you're saying that we shouln't spam slurm logs with this fake error. 
Agreed

awennersteen (approved):
lgtm, thanks!

src/pasqal/cloud.rs
badtst: Yes, this error log would creep in my user script output from the Spank plugin instantiation

dependencies/pasqal_cloud_client/src/client.rs
```diff
@@ -324,6 +327,12 @@ impl Client {
         self.handle_request(resp).await
     }
 
+    // Unauthenticated GET request for public endpoints
+    pub(crate) async fn get_raw<T: DeserializeOwned>(&mut self, url: &str) -> Result<T> {
```
MatthieuMoreau0: nit: 
```suggestion
    pub(crate) async fn get_unauthenticated<T: DeserializeOwned>(&mut self, url: &str) -> Result<T> {
```

MatthieuMoreau0 (approved):
Tested locally, works like a charm, thanks!

dependencies/pasqal_cloud_client/src/client.rs
badtst: Yes makes more sense thanks ! Done

PR #142, opened 2026-06-09
Add ResourceProvider for dynamic backend resource discovery

### LINKED ISSUE #52: [DRAFT] EPIC: Resource Discovery

Background: In the current implementation, HPC users must specify the quantum computer resource name (backend name) using the `--qpu` argument for `sbatch` or `srun`. This works well when HPC users are familiar with all quantum computers available on the cluster and know exactly which device they want to use. However, there are also use cases where users want to select any backend that satisfies certain constraints—such as topology, technology, or the number of qubits—rather than specifying a particular backend name.

Allow HPC users to specify certain qubit count and other constraints to allocate QPU resource for use, instead of specifying quantum computer name.

`QuantumResourceProvider(`new trait) should have corresponding interface like below.

`resources (self, min_qubits: int32, constraints: &str) -> Result<Vec<QuantumResource>>;`

Each QRMI implementation implements this function and interprets the constraint string—representing resource‑unique constraints—specified in constraint, and returns to the caller a list of `QuantumResource` instances that satisfy those conditions.

Similar function should be created for Python and C binding as well. C binding will be used by spank plugin. Python binding will be used to implement QRMI python library such as Qiskit provider like library.

QiskitRuntimeService provides the following 2 methods:

- [backends()](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/qiskit-runtime-service#backends) ... returns a list of quantum backends with filtering by:
  - Minimum number of qubits the backend has to have.
  - Filter by whether the backend supports dynamic circuits.
  - More complex filters, such as lambda functions

- [least_busy()](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/qiskit-runtime-service#least_busy) ... Return the least busy available backend.

T.B.D

At least, we need to enhance 
- QRMI API document(Rust/C/Python)
- API examples
- Slurm job examples
- qrmi primitives python library etc.

ohtanim:
### Overview
In our current implementation, system administrators are expected to create a list of resource definitions in qrmi_config.json, where each entry is a block containing the exact backend name, resource type, and the list of environment variables needed to access it. However, this approach has maintainability challenges: the qrmi_config.json on compute nodes needs to be updated whenever a new backend is added, and similar resource definitions end up being repeated throughout the file.

This PR introduces ResourceProvider, a new abstraction layer that enables dynamic backend discovery for quantum resource providers. Users can now enumerate available backends from a provider (e.g. IBM Qiskit Runtime Service, IBM Quantum System) at runtime, rather than listing them statically in the config file.

I will not introduce the modifications of the existing `QuantumResource` implementation - No impacts to your code. The timing of the implementation will also be left to each vendor. As usual, we may revisit and revise the interface based on actual usage.

#### New trait: ResourceProvider
A new `ResourceProvider` trait is added in `src/resource_provider.rs` with two methods:

- `resources(filters: Option<String>)` — returns available backends sorted by queue_length ascending (least busy first)
- `least_busy(filters: Option<String>)` — convenience method that returns the first element of resources()

A `create_provider()` factory function is also provided, which selects the correct provider implementation based on ResourceType.

#### Config file extension: is_dynamic
ResourceDef in `qrmi_config.json` now supports an `is_dynamic` field (default: `false`). When true, the entry acts as a provider profile for dynamic backend discovery rather than a static resource definition.

```
{
    "resources": [
        {
            "name": "ibm_inst1",
            "type": "qiskit-runtime-service",
            "is_dynamic": true,
            "environment": {
                "QRMI_IBM_QRS_ENDPOINT": "https://quantum.cloud.ibm.com/api/v1",
                "QRMI_IBM_QRS_IAM_APIKEY": "<your_api_key>",
                ...
            }
        }
    ]
}
```

#### Provider implementations

- `IBMQiskitRuntimeServiceProvider `— discovers backends via `list_backends()` and `get_backend_configuration()`, sorted by `queue_length`
- `IBMQuantumSystemProvider` — discovers backends via `list_backends()` and `get_backend_configuration()`, returned in server order (no queue)

#### Filter(constraints) string format
Both providers support a common filter string of the form key=value&key=value

| Key | Description | Default |
| --- |  --- | --- |
| num_qubits=<N> | Only backends with n_qubits >= N | none |
| max_shots=<N> | Only backends with max_shots >= N | none |
| name=<glob> | Glob pattern match on backend name | none |
| is_simulator=<bool> | Include/exclude simulators | false |
| status=online | Only online backends | none |

#### Shared BackendConfiguration model (For IBM only)
A shared `BackendConfiguration` model is introduced at `src/ibm/models/backend_configuration.rs`, used by both provider implementations. This avoids duplication and ensures consistent deserialization behavior (including handling of null fields and revision as string or integer).

#### Python bindings

- `ResourceProvider(resource_type, environment)` — creates a provider
- `Config.load(path) `— loads config file, returns `Config` with `resource_map: dict[str, ResourceDef]`
- `ResourceDef` — exposes `name`, `resource_type`, `is_dynamic`, `environment`

#### C bindings

- `qrmi_provider_new(resource_type, environments) `— creates a provider
- `qrmi_provider_resources(provider, filter, &resources)` — returns QrmiQuantumResources
- `qrmi_provider_resources_free(&resources)` — frees the resources struct
- `qrmi_provider_least_busy(provider, filter, &resource_out)` — returns the least busy resource
- `qrmi_provider_free(provider)` — frees the provider

#### Unified examples
New unified examples `examples/qrmi/*/providers/` replace the separate per-provider examples. The resource type is read from the config file, so no code changes are needed when switching between providers.

Rust:
`cargo run --example qrmi-example-providers -- /etc/slurm/qrmi_config.json ibm_inst1 -f "num_qubits=127"
`

Python:
`python example.py /etc/slurm/qrmi_config.json ibm_inst1 --filters "num_qubits=127&name=ibm_*"
`

C:
`./build/providers /etc/slurm/qrmi_config.json ibm_inst1 "num_qubits=127&max_shots=10000"`

#### Backward compatibility
Existing `qrmi_config.json` files without `is_dynamic` continue to work unchanged — the field defaults to false. All existing C and Python APIs are unaffected.

## Checklist ✅

## Ticket
- [x] Fixes #52 
- [ ] Is Part of #


src/cext.rs
```diff
@@ -1481,3 +1532,368 @@ pub unsafe extern "C" fn qrmi_resource_metadata_keys(
...
+
+// ---------------------------------------------------------------------------
+// ResourceProvider C bindings
+// ---------------------------------------------------------------------------
+
+/// IBM Qiskit Runtime Service resource provider handle.
+pub struct ResourceProvider {
+    inner: Box<dyn crate::ResourceProvider>,
+    runtime: Arc<tokio::runtime::Runtime>,
+}
```
awennersteen:
```suggestion
/// Resource provider handle.
pub struct ResourceProvider {
    inner: Box<dyn crate::ResourceProvider>,
    runtime: Arc<tokio::runtime::Runtime>,
}
```

right? nothing specific to IBM Qiskit Runtime Service about it?

ohtanim: @awennersteen Good catch. You're right. I will fix the comment soon.

src/cext.rs
```diff
@@ -1481,3 +1532,368 @@ pub unsafe extern "C" fn qrmi_resource_metadata_keys(
...
+/// * `provider` must have been returned by a previous call to qrmi_provider_new().
+/// * `filters` may be NULL (no filtering) or a valid nul-terminated C string.
+/// * `resources_out` must be non-null and point to a zero-initialized QrmiQuantumResources.
+///
+/// # Filter string format
+///
+/// `key=value` pairs joined by `&`. Supported keys:
+/// - `num_qubits=<N>`      — only backends with qubits >= N
+/// - `name=<glob>`         — glob pattern match on backend name
+/// - `is_simulator=<bool>` — include/exclude simulators (default: false)
```
awennersteen: I'm confused by this name `/// - `is_simulator=<bool>` — include/exclude simulators (default: false)`.
Should it instead be `include_simulators`?

```suggestion
/// - `include_simulators=<bool>` — include/exclude numerical simulators / emulators (default: false)
```

Simulator is a overloaded term in quantum, so I'm just adding the word "emulators" there too to show up when grepping the codebase

src/cext.rs
ohtanim: Fixed.

src/pyext.rs
```diff
@@ -226,6 +239,232 @@ impl PyQuantumResource {
...
+    /// Returns available quantum resources, optionally filtered.
+    ///
+    /// # Arguments
+    ///
+    /// * `filters` - Filter string of the form `key=value&key=value`, or `None`.
+    ///
+    /// Supported keys:
+    /// - `num_qubits=<N>`      — only backends with `qubits >= N`
+    /// - `name=<glob>`         — only backends whose name matches the glob pattern
+    /// - `is_simulator=<bool>` — include/exclude simulators (default: `false`)
```
awennersteen: Same comment as in cext

```suggestion
    /// - `include_simulators=<bool>` — include/exclude numerical simulators / emulators (default: `false`)
```

examples/qrmi/rust/Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
     "qrmi_config",
     "alice_bob_felis",
     "iqm_server",
+    "providers",
```
awennersteen: I'm wondering if just `providers` is a good name. `dynamic_providers`? or `resource_providers` as used more in the code

src/cext.rs
ohtanim: Ok. I will change to "includes_simulator".

Regarding "simulator" or "emulator", I think this is depending on each vendor's terminology. In case of IBM Quantum platform, we use "simulator" in the API description.

examples/qrmi/rust/Cargo.toml
ohtanim: Good suggestion. I will rename to "resource_providers". Thanks,

src/cext.rs
awennersteen: yes, i'm happy to go with simulator.

awennersteen:
Overall, lgtm to me, I will read over again a little bit. I'm not going to review in great detail the IBM specific code (but will read), and I can't test it.

I think it's the right choice of abstraction. Also since scheduler constructs like GRES might be hard to define with these dynamic resources (and corresponding in other schedulers). 
It looks extensible to me, I don't see any issues. I will try to do it for our cloud service as a validation asap - but no need to hold back for it

examples/qrmi/rust/Cargo.toml
ohtanim: Changed the directory name to `resource_provider` for rust/c/python examples.

ohtanim:
> Overall, lgtm to me, I will read over again a little bit. I'm not going to review in great detail the IBM specific code (but will read), and I can't test it.
> 
> I think it's the right choice of abstraction. Also since scheduler constructs like GRES might be hard to define with these dynamic resources (and corresponding in other schedulers). It looks extensible to me, I don't see any issues. I will try to do it for our cloud service as a validation asap - but no need to hold back for it

@awennersteen Thank you. Regarding GRES, I believe there are various implementations. At the very least, if it enables resource specifications like -`-gres=qpu:ibm_127q:2`, this implementation should be usable. I will create a separate PR for the SPANK plugin.

src/pyext.rs
ohtanim: I'm going to remove this source code comment. Each vendor will define the constraints(filter string), so those descriptions should not be placed here.

awennersteen: ah gotcha, I didn't even realise this would be fully up to the resource provider implementation for each vendor. thanks

ohtanim: Updated the source code comment like below.

```
/// Filter specifications (constraints) are defined by each resource provider's implementation.
/// Results are expected to be sorted in least-busy order.
```

src/cext.rs
ohtanim: Oops.

I will fix this source code comments. Same to src/pyext.rs.
Current source code comments are just describing IBM implementation. They should not described in common code. I will change the comment descriptions like being done in src/pyext.rs.

ohtanim:
@awennersteen @OkuyanBoga @yoonho @machjn In this PR, the definition of supported filter string (= constraints) key/value pairs is left to each provider. However, as vendor implementations progress, we expect common constraints to emerge. At the very least, num_qubits and name will likely become common.
As part of the QRMI open-source development policy, we first release a beta version, have users and projects try it out, and then incorporate their feedback. This PR is the first step in that process.

examples/qrmi/python/resource_providers/README.md
```diff
@@ -0,0 +1,34 @@
...
+  config_file        Path to qrmi_config.json
+  resource_name      Name of the dynamic resource definition (is_dynamic=true)
+
+options:
+  -h, --help         show this help message and exit
+  --filters FILTERS  Optional filter string e.g. 'num_qubits=127&name=ibm_*'
+```
+For example,
+```shell-session
+python example.py /etc/slurm/qrmi_config.json ibm_inst1 --filter "num_qubits=127&max_shots=10000"
```
OkuyanBoga:
```suggestion
python example.py /etc/slurm/qrmi_config.json ibm_inst1 --filters "num_qubits=127&max_shots=10000"
```

examples/qrmi/c/resource_providers/src/providers.c
```diff
@@ -0,0 +1,96 @@
...
+    qrmi_string_free((char *)err);
+    return EXIT_FAILURE;
+  }
+
+  QrmiQuantumResources resources = {0};
+  QrmiReturnCode rc = qrmi_provider_resources(provider, filter, &resources);
+  if (rc == QRMI_RETURN_CODE_SUCCESS) {
+    for (size_t i = 0; i < resources.length; i++) {
+      char *id = NULL;
+      qrmi_resource_id(resources.resources[i], &id);
```
OkuyanBoga, line 73: In case of fail, I think ID might be NULL, and might cause issues. I would suggest doing something like
```suggestion
      if (qrmi_resource_id(resources.resources[i], &id) != QRMI_RETURN_CODE_SUCCESS || id == NULL) {
    continue;
}
```

src/pyext.rs
```diff
@@ -226,6 +239,227 @@ impl PyQuantumResource {
...
+/// ```python
+/// from qrmi import Config, ResourceProvider
+///
+/// config = Config.load("/path/to/qrmi_config.json")
+///
+/// # Iterate over all resource definitions
+/// for name, resource_def in config.resource_map.items():
+///     print(f"{name}: is_dynamic={resource_def.is_dynamic}")
+///     if resource_def.is_dynamic:
+///         provider = ResourceProvider(resource_def)
```
OkuyanBoga: I think example config is not updated
```suggestion
///         provider = ResourceProvider(resource_def.resource_type, resource_def.environment)
```

src/resource_provider.rs
```diff
@@ -0,0 +1,117 @@
...
+///         println!("least busy: {}", r.resource_id().await?);
+///     }
+///     Ok(())
+/// }
+/// ```
+#[async_trait]
+pub trait ResourceProvider: Send + Sync {
+    /// Returns a list of available quantum resources, optionally filtered.
+    ///
+    /// Results are sorted by `queue_length` ascending (least busy first).
```
OkuyanBoga, line 55: Is this correct for other vendors? I think we should ask providers to sort:

```suggestion
    /// Providers should return resources in preferred order; queue-based providers
    /// should sort by `queue_length` ascending.
```

src/ibm/quantum_system_provider/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+                        );
+                        return None;
+                    }
+                };
+
+                if !filter.matches_config(&config) {
+                    return None;
+                }
+
+                self.inject_backend_env(&name);
```
OkuyanBoga, line 180: I am not sure but could this create an issue regarding leaking credentials through environment state?

src/cext.rs
```diff
@@ -1481,3 +1532,364 @@ pub unsafe extern "C" fn qrmi_resource_metadata_keys(
...
+            }
+        },
+        _ => {
+            _set_last_error("Unsupported resource type for dynamic resource discovery".to_string());
+            return std::ptr::null_mut();
+        }
+    };
+    Box::into_raw(Box::new(ResourceProvider {
+        inner: provider,
+        runtime: Arc::new(tokio::runtime::Runtime::new().unwrap()),
```
OkuyanBoga, line 1621: I am not sure but '.unwrap' generally worries me when using c_api. I think this might cause an issue through the C API if runtime creation fails. I suggest returning NULL and maybe setting last error too instead of unwrapping, just in case.

OkuyanBoga (changes requested):

examples/qrmi/python/resource_providers/README.md
OkuyanBoga: This is a simple fix for the Readme, I applied the fix directly.

OkuyanBoga:
Thank you @ohtanim, overall looks good to me. I made a quick review with some comments and suggestions.

yoonho (approved):
Since we are beginning to integrate with different resource managers (and learning from it), I really like the idea of prototyping something quickly that does not affect existing deployments. We can certainly try this with our LSF/QRMI integration, and I am sure that Vanessa will want to try it with her Fluxion/QRMI integration. I do not know if we have the right vendor separation (eg, can Pascal and A&B provide a dynamic resource too?), but I am sure we will find out.

src/ibm/quantum_system_provider/mod.rs
ohtanim: @OkuyanBoga It will continue to operate in an environment-variable–driven manner as before. As an approach to avoid including credentials in environment variables, the use of an auth proxy has been discussed. However, that is a topic to be handled in a separate PR, and I don’t think it should be included as a review comment on this PR. Make sense?

src/pyext.rs
ohtanim: Fixed.

awennersteen (approved):
lgtm. great work.

I have some doubts going forward (boundaries, shared interfaces, future developments etc) but I think it's best discussed after someone tries to extend it

I found a couple more typos for docstrings that are upcoming

src/cext.rs
```diff
@@ -1481,3 +1532,364 @@ pub unsafe extern "C" fn qrmi_resource_metadata_keys(
...
+/// # Safety
+///
+/// * `ptr` must have been returned by a previous call to qrmi_provider_new().
+///
+/// # Example
+///
+/// @code
+///   QrmiResourceProvider *provider = qrmi_provider_new();
+///   if (provider != NULL) {
+///     qrmi_provider_free(provider);
```
awennersteen: nit: 

Should be 
`qrmi_provider_new(resource_type, environment)`
right?

examples/qrmi/c/resource_providers/src/providers.c
ohtanim: This cannot happen. In the first place, the resource ID represents a discovered resource name such as "ibm_kingston". A provider would never return a resource id with a NULL value for this, and if it did, that would be considered an error on the provider side.

I would keep existing implementation because segmentation fault will notify us such provider side issue.

Closing.

src/cext.rs
ohtanim: Good catch.

OkuyanBoga (approved):

src/cext.rs
ohtanim: Fixed.

ohtanim:
> Since we are beginning to integrate with different resource managers (and learning from it), I really like the idea of prototyping something quickly that does not affect existing deployments. We can certainly try this with our LSF/QRMI integration, and I am sure that Vanessa will want to try it with her Fluxion/QRMI integration. I do not know if we have the right vendor separation (eg, can Pascal and A&B provide a dynamic resource too?), but I am sure we will find out.

Thanks. I plan to implement for IQM soon.

PR #144, opened 2026-06-11
[PASQAL] Refactor Pasqal client and cloud resource. Add service accou…

awennersteen:
## Description

This PR Makes the PASQAL_CONFIG_ROOT settable in qrmi_config.json, enabling sysadmins to to set it to e.g. `/work/$USER/`, for systems that don't mount `/home/` on worker nodes.

It also adds service accounts to the auth methods, and performs much needed refactor of code into logically separated files.

## Description of Change

- Add support for client_id / client_secret in .pasqal/config for service-account authentication.
- Factor Pasqal Cloud auth code out of the client implementation into auth/auth.rs.
- Move Pasqal Cloud API response/request models under models/.
- Add config helpers for Pasqal Cloud env/config resolution  into cloud_config.rs.
- Support PASQAL_CONFIG_ROOT config qrmi_config.json, including runtime expansion of ~, $VAR, and ${VAR}.
- Split Pasqal Cloud client endpoint methods (get_batch_) from HTTP helpers (get) in separate impl Client blocks.

### Credential Resolution

Pasqal Cloud credentials can now come from:

- Explicit env vars
- User .pasqal/config
- Cluster qrmi_config.json defaults where applicable

Credential type priority is:

username/password > client_id/client_secret > token

Within a given credential type, environment variables still take precedence over config-file values.

Supported .pasqal/config keys now include:

username=...
password=...
client_id=...
client_secret=...
token=...
project_id=...
auth_endpoint=...

PASQAL_CONFIG_ROOT values may be provided by admins in qrmi_config.json; QRMI expands ~, $VAR, and ${VAR} at runtime, then reads:

<expanded-root>/.pasqal/config

Example:

{
  "name": "EMU_FREE",
  "type": "pasqal-cloud",
  "environment": {
    "PASQAL_CONFIG_ROOT": "/work/$USER"
  }
}

resolves for user aleks to:

/work/aleks/.pasqal/config

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #


awennersteen:
@ohtanim and @OkuyanBoga no need to review this - it's purely internal changes to pasqal code base

dependencies/pasqal_cloud_client/src/auth/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+
+        if resp.status().is_success() {
+            let token: AuthTokenResponse = resp.json().await?;
+            Ok(token.access_token)
+        } else {
+            let status = resp.status();
+            let body = resp.text().await.unwrap_or_default();
+            bail!("Token request failed: {} {}", status, body);
+        }
+    }
```
awennersteen: This code is new - I opted for a new function because at the time it looked clearer but now I'm not so sure

dependencies/pasqal_cloud_client/src/client.rs
```diff
@@ -317,6 +175,15 @@ impl Client {
             Ok(specs_string)
         }
     }
+}
+
+impl Client {
+    fn build_http_client() -> Result<reqwest_middleware::ClientWithMiddleware> {
```
awennersteen, line 188: This second `impl Client` block defines all the standard methods, just to separate the "business logic" client methods from above

src/pasqal/cloud.rs
badtst, line 70: I think now that the Spank plugin has no need for authentication to check `is_accessible` or `acquire`, we should be able to let the spank plugin create a pasqal-cloud resource without any authentication/project_id. 

That means that we should be able to have users configure `project_id` and auth inside their `.pasqal/config` file while having an empty `environment` in `qrmi_config.json` for the pasqal-cloud resources. That doesn't work for the moment since a `project_id` value is mandatory for the creation of a cloud client .

Lmk if it isn't clear

badtst: For example if I set my `qrmi_config.json` like: 

```json
    {
      "name": "EMU_FREE",
      "type": "pasqal-cloud",
      "environment": {
        "QRMI_PASQAL_CLOUD_PROJECT_ID": "blabla" <-- Needed by spank plugin emu_free instantiation but no requests actually needs it
      }
    }
```

And then in my sbatch script 
```bash
unset EMU_FREE_QRMI_PASQAL_CLOUD_PROJECT_ID <-- Would resolve to "blabla"
python /path/to/pulser_qrmi.py
```

Works if I have a correct `poject_id` defined in `PASQAL_CONFIG_ROOT/.pasqal/config` for the user QRMI.

I'd like to be able to have my `qrmi_config.json` like:

```json
    {
      "name": "EMU_FREE",
      "type": "pasqal-cloud",
      "environment": {}
    }
```

And not fail because the Spank plugin does not have a `project_id` to instantiate a EMU_FREE QRMI resource

badtst: Or maybe that's not a desired behavior but since we allow users to overwrite the `project_id` anyways why not ?

dependencies/pasqal_cloud_client/src/client.rs
```diff
@@ -13,38 +13,25 @@
...
 
-use crate::models::batch::JobStatus;
-use crate::models::device::DeviceType;
-use base64::engine::general_purpose::URL_SAFE_NO_PAD;
-use base64::Engine as _;
+use crate::models::{
+    Batch, CancelBatchResponseData, CreateBatchResponseData, CreateCudaqJobResponseData,
+    DeviceType, GetBatchResponseData, GetCudaqJobResponseData, GetDeviceResponseData,
+    GetDeviceSpecsResponseData, GetJobResponseData, Job, JobResult, Response,
+};
```
badtst: nit: might be more readable to do something like 
```suggestion
use crate::models;
```
And then prefix the models when using them like `model::Batch`

src/pasqal/cloud.rs
awennersteen: great idea - it is a natural continuation of letting users supply their own auth info in `.pasqal/config`.

It goes a bit against the spirit we originally had in mind, but it wasn't possible to adapt our platform enough in that direction anyways. So I think we should do as you say

badtst: Okay, maybe in another PR or do you think here is the right place ?

dependencies/pasqal_cloud_client/src/auth/auth.rs
MatthieuMoreau0: this looks alright to me

dependencies/pasqal_cloud_client/src/auth/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    fn normalize_auth_endpoint(auth_endpoint: &str, auth_method: &str) -> Result<String> {
```
MatthieuMoreau0: nit: auth_method argument doesn't change anything but the log, we don't really need it

src/pasqal/cloud.rs
```diff
@@ -244,15 +75,19 @@ impl PasqalCloud {
             builder.with_base_url(base_url);
         }
 
-        // Preference order for credentials: explicit username/password in env > config > direct token.
+        // Preference order: username/password > service account credentials > direct token.
```
MatthieuMoreau0: This preference order is already implemented in the `ensure_authenticated` method of the client. So here we could simply pass everything from the config and let the client decide what to use

src/pasqal/cloud_config.rs
```diff
@@ -0,0 +1,339 @@
...
+        .iter()
+        .find_map(|path| fs::read_to_string(path).ok())
+    {
+        Some(content) => content,
+        None => {
+            for path in configured_config_root_paths {
+                warn!(
+                    "Pasqal config root is set but config file was not found: {}",
+                    path.display()
+                );
```
MatthieuMoreau0, line 224: nit: imo we could print the full path instead of the just the root, just in case user misspell .pasqal/config part.

another nit: a single warning with the list of paths we checked is a bit less spammy than 1 line per path imo

MatthieuMoreau0 (commented):
Overall this looks good, minor comments

src/pasqal/cloud.rs
MatthieuMoreau0: In the current implem, the project_id in the config file of the user is always ignored because the env variable defined by the spank plugin takes precedence

MatthieuMoreau0:
Something I find a bit odd with this implementation is the role of the PASQAL_CONFIG_ROOT in the qrmi_config.json. Here it defines both the path of the config file that the spank plugin should use as well as the path where end users should put their config file. Is there no realistic scenario where these do not match? If yes, maybe we need to define two separate keys in the qrmi_config.json file

src/pasqal/cloud_config.rs
```diff
@@ -0,0 +1,339 @@
...
+            .ok()
+            .filter(|v| !v.trim().is_empty())
+            .or(self.project_id.clone().filter(|v| !v.trim().is_empty()))
+            .or(read_qrmi_config_env_value(
+                backend_name,
+                "QRMI_PASQAL_CLOUD_PROJECT_ID",
+            ))
+            .ok_or_else(|| {
+                anyhow!(
+                    "{project_id_var} is not set and no project_id was found in ~/.pasqal/config or /etc/slurm/qrmi_config.json"
```
MatthieuMoreau0: nit: this error message is a bit misleading now that the config file path can be set by user

src/pasqal/cloud_config.rs
```diff
@@ -0,0 +1,339 @@
...
+        .filter(|v| !v.trim().is_empty())
+        .or(cfg.username.clone().filter(|v| !v.trim().is_empty()));
+    let password = env::var("PASQAL_PASSWORD")
+        .ok()
+        .filter(|v| !v.trim().is_empty())
+        .or(cfg.password.clone().filter(|v| !v.trim().is_empty()));
+    (username, password)
+}
+
+pub(crate) fn resolve_pasqal_service_account_credentials(
```
MatthieuMoreau0: Why not define getter (like for project_id and auth_token) on the config instead of these functions?

src/pasqal/cloud_config.rs
```diff
@@ -0,0 +1,339 @@
...
+            path.push("config");
+            config_path_candidates.push(path);
+        }
+    }
+
+    let content = match config_path_candidates
+        .iter()
+        .find_map(|path| fs::read_to_string(path).ok())
+    {
+        Some(content) => content,
```
MatthieuMoreau0: I think it would be nice to print the path that was chosen to read the config (since there is a bunch of options, it will make debugging easier)

src/pasqal/cloud_config.rs
```diff
@@ -0,0 +1,339 @@
...
+        }
+    }
+
+    Ok(config)
+}
+
+pub(crate) fn read_qrmi_config_env_value(backend_name: &str, key: &str) -> Option<String> {
+    let content = fs::read_to_string("/etc/slurm/qrmi_config.json").ok()?;
+    read_qrmi_config_env_value_from_content(&content, backend_name, key)
+}
```
MatthieuMoreau0: I thought the spank plugin was responsible for reading the qrmi_config.json file and inject the env variables? then the qrmi only checks the env variable directly

Is that not the case?

There are a couple issues with this:
- end user will not have rights to access the file
- path is hard-coded whereas in the sopank plugin the location is chosen by the admin so the file may not be in /etc/slurm

Looks like this is not new, I missed this on a previous review.

awennersteen: yeah, I think this came also when looking either at CUDA-Q or Grid Engine integration early stage. But I tested and it can be removed.

awennersteen:
>Something I find a bit odd with this implementation is the role of the PASQAL_CONFIG_ROOT in the qrmi_config.json. Here it defines both the path of the config file that the spank plugin should use as well as the path where end users should put their config file. Is there no realistic scenario where these do not match? If yes, maybe we need to define two separate keys in the qrmi_config.json file


I'm not sure the current implementation and ideas can support well mixing those. it wouldhave to be with env var overrides. I'm not sure thats easy to implement. 
The goal here was simply to allow an admin to define any patterns that might be of the final file location - as controlled by them. I've seen this at quite a few clusters.

src/pasqal/cloud_config.rs
MatthieuMoreau0: If we can do without it for CUDA-Q and Grid Engine, then I'd advocate for removing this yeah for the reasons mentionned above. If we are unsure or we need to tweak cuda-q/grid engine implem let's do that as part of another MR, this is out of scope

dependencies/pasqal_cloud_client/src/auth/auth.rs
awennersteen: fixed

src/pasqal/cloud.rs
awennersteen: yep, thanks

src/pasqal/cloud_config.rs
awennersteen: I now print the read file in debug, and collect all the paths to be logged at once. And log the full path.

src/pasqal/cloud_config.rs
```diff
@@ -184,56 +184,36 @@ pub(crate) fn expand_env_vars(value: &str) -> Result<String> {
...
-    }
+pub(crate) fn read_pasqal_config() -> Result<PasqalConfig> {
+    let config_root_path = match env::var("PASQAL_CONFIG_ROOT").ok() {
+        Some(config_root) => pasqal_config_path_from_root(&config_root)?,
+        None => None,
+    };
+    let home_config_path = env::var("HOME")
+        .ok()
+        .filter(|home| !home.trim().is_empty())
+        .map(|home| PathBuf::from(home).join(".pasqal").join("config"));
```
badtst: nit: why not reuse the `pasqal_config_path_from_root` function here ?

src/pasqal/cloud.rs
```diff
@@ -61,8 +58,8 @@ impl PasqalCloud {
             backend_name
         );
 
-        let cfg = PasqalConfig::read(backend_name)?;
-        let project_id = cfg.project_id(backend_name)?;
+        let cfg = PasqalConfig::read()?;
+        let project_id = cfg.project_id(backend_name).unwrap_or_default();
```
badtst, line 63: nit: setting the default value like that here is fine, it could also have been set in the `ClientBuilder` the same way it is done for the token or auth_endpoint default values

src/pasqal/cloud_config.rs
```diff
@@ -184,56 +184,36 @@ pub(crate) fn expand_env_vars(value: &str) -> Result<String> {
...
 
-    for backend_config_root_var in [
-        format!("{backend_name}_QRMI_PASQAL_CONFIG_ROOT"),
-        format!("{backend_name}_PASQAL_CONFIG_ROOT"),
-    ] {
-        if let Ok(config_root) = env::var(&backend_config_root_var) {
-            if let Some(path) = pasqal_config_path_from_root(&config_root)? {
```
badtst: So we get rid of the possibility to configure PASQAL_CONFIG_ROOT in the qrmi_config.json. Fine by me

badtst (approved):
lgtm, thanks !

MatthieuMoreau0 (approved):
Tested and works locally, thanks!

src/pasqal/cloud.rs
```diff
@@ -244,15 +72,29 @@ impl PasqalCloud {
...
+        if project_id.is_empty() {
+            debug!(
+                "No Pasqal Cloud project_id configured for backend '{}'; unauthenticated operations can still be used",
+                backend_name
+            );
+        }
+        if !has_user_credentials && !has_service_account_credentials && !has_auth_token {
             debug!(
-                "No Pasqal Cloud auth details configured for backend '{}': expected PASQAL_USERNAME/PASQAL_PASSWORD, ~/.pasqal/config credentials, or token",
+                "No Pasqal Cloud auth details configured for backend '{}': expected PASQAL_USERNAME/PASQAL_PASSWORD, ~/.pasqal/config credentials, client_id/client_secret, or token",
```
MatthieuMoreau0, line 98: Out of curiosity are you able to see these logs in slurmd? I assume it's a matter of logger config, but by default I don't see them

ohtanim: > Out of curiosity are you able to see these logs in slurmd? I assume it's a matter of logger config, but by default I don't see them

@MatthieuMoreau0 Can you try ?
https://github.com/qiskit-community/spank-plugins/tree/main/plugins/spank_qrmi#logging

MatthieuMoreau0: By default in our test cluster I already see debug logs in slurmd for the spank plugin but not the logs from the QRMI.

Thank you for the link, I tried to set the `--verbose` flag when running sbatch and updating slurm.conf to `SlurmdDebug=4` but I still don't see the logs from the QRMI

I also tried changing the log level in the QRMI to info, warn or error and still couldn't see them appearing in slurmd

ohtanim: I will check in parallel.

awennersteen: I think the QRMI debug logs go into the job log. e.g. `job_10.out`, for example I have:
```
[2026-06-18T17:45:07Z DEBUG qrmi::pasqal::cloud] Initializing PasqalCloud QRMI for backend 'EMU_FREE'
[2026-06-18T17:45:07Z DEBUG qrmi::pasqal::cloud_config] Reading Pasqal config file: /shared/qrmi-slurm-test-config/.pasqal/config
[2026-06-18T17:45:07Z DEBUG qrmi::pasqal::cloud] Build PasqalCloud client for backend 'EMU_FREE'
[2026-06-18T17:45:07Z DEBUG pasqal_cloud_api::client] Initialize Client (project_id set: true, auth_token set: true, username/password set: false/false, client_id/client_secret set: true/true)
[2026-06-18T17:45:07Z DEBUG reqwest::connect] starting new connection: https://apis.pasqal.cloud/
```

But it is true when you say it that this might be better routed elsewhere

MatthieuMoreau0: Ideally it should appear in both since both the spank plugin and the user script instantiate the QRMI

ohtanim: I agree.

I took a quick look at Slurm’s implementation, and it appears that the stdout and stderr of the slurmd process are simply discarded. If we were to handle this, it seems the only option would be to write to a QRMI-specific log file.

Also, our current SPANK plugin does not set the RUST_LOG environment variable for the slurmd process, so the SlurmdDebug configuration is not taking effect.

src/pasqal/cloud_config.rs
awennersteen: It may have been lost in one particular commit, but it is now back. it is picked up as
```
<backend_name>_PASQAL_CONFIG_ROOT
```
env var, which if we had `PASQAL_CONFIG_ROOT` to `qrmi_config.json` is automatically exported by the scheduer plugin.

PR #155, opened 2026-07-03
[PASQAL] Add pasqal clients retry policy

sarjevane:
## Description of Change

- Add PASQAL's `pasqal_common` package containing common retry logic
- Apply retry policy for both `pasqal-local` and `pasqal-cloud`
- Disable retry policy using environment variable (for e2e tests) using `QRMI_PASQAL_RETRIES_DISABLED=1`
- Handle two classes of retries:
  - rate-limiting (423/429) with handling of `Retry-After`, defaults to jitter / exponential backoff, max 5 minutes
  - transient errors (5xx) - with jitter and exponential backoff, max 2 minutes

## Checklist ✅

### DIFF
--- a/dependencies/pasqal_common/src/retry.rs
+++ b/dependencies/pasqal_common/src/retry.rs
@@ -14,24 +14,29 @@
 //! Retries are split into two tiers:
 //!
 //! * ordinary transient failures (5xx, timeouts, connection errors), handled by
-//!   the stock [`RetryTransientMiddleware`] with an exponential backoff, and
+//!   the stock [`RetryTransientMiddleware`] with an exponential backoff that
+//!   starts at 1s, doubles, and is capped at 5s per wait, and
 //! * rate-limit / locked responses (HTTP 429 and 423), handled by
-//!   [`RetryAfterMiddleware`], which honors the server's `Retry-After` header
-//!   when present and otherwise falls back to the same exponential backoff.
+//!   [`RetryAfterMiddleware`]. When the server sends a `Retry-After` header we
+//!   trust it and retry exactly once, since the server has told us when it will
+//!   be ready and guessing again afterwards only adds load. With no such header
+//!   we fall back to an exponential backoff that starts at 5s, doubles, and is
+//!   capped at 30s per wait — a rate-limited backend wants more breathing room
+//!   between polls than a merely flaky one.
 //!
-//! Both tiers are bounded by wall-clock time rather than a retry count, so a
-//! request never spends longer than its tier's `max_total_duration`
-//! retrying, no matter how many attempts that allows. The rate-limit tier gets
-//! a larger per-attempt ceiling and a larger total budget, since a busy backend
-//! wants to be given more room between polls.
+//! Both tiers are bounded by a retry *count* rather than a wall-clock budget,
+//! and share the same count so that a caller has a single number to reason
+//! about (see [`with_retry`]). A count of zero disables both tiers, including
+//! the `Retry-After`-guided retry.
 
 use std::time::{Duration, SystemTime};
 
+use anyhow::anyhow;
 use async_trait::async_trait;
 use http::Extensions;
 use log::debug;
 use reqwest_middleware::{ClientBuilder, Middleware, Next};
-use reqwest_retry::policies::{ExponentialBackoff, ExponentialBackoffTimed};
+use reqwest_retry::policies::ExponentialBackoff;
 use reqwest_retry::{
     default_on_request_failure, default_on_request_success, Jitter, RetryDecision, RetryPolicy,
     RetryTransientMiddleware, Retryable, RetryableStrategy,
@@ -39,6 +44,17 @@ use reqwest_retry::{
 
 const HTTP_TOO_MANY_REQUESTS: u16 = 429;
 const HTTP_LOCKED: u16 = 423;
+
+/// How many times a request is retried, absent an explicit choice by the caller.
+pub const DEFAULT_MAX_RETRIES: u32 = 5;
+
+/// Longest wait a server's `Retry-After` header can ask of us.
+///
+/// The header is honored as sent, up to this ceiling: a backend that asks us to
+/// come back in an hour is telling us it is not coming back within the lifetime
+/// of any request we care about, so the request fails with an error naming the
+/// wait the server asked for instead of sleeping on it.
+const RETRY_AFTER_MAX_WAIT: Duration = Duration::from_secs(300);
 
 /// Backoff parameters for a single class of retryable responses.
 #[derive(Debug, Clone)]
@@ -51,14 +67,13 @@ struct BackoffConfig {
     base: u32,
     /// Whether to apply bounded jitter to smooth out retry storms.
     jitter: bool,
-    /// Hard cap on the total wall-clock time spent retrying a request. Once
-    /// this budget is exhausted the request fails, regardless of attempt count.
-    max_total_duration: Duration,
+    /// How many times a failed request is retried before it is given up on.
+    max_retries: u32,
 }
 
 impl BackoffConfig {
-    /// Compile these parameters into a time-bounded exponential backoff policy.
-    fn build_policy(&self) -> ExponentialBackoffTimed {
+    /// Compile these parameters into a count-bounded exponential backoff policy.
+    fn build_policy(&self) -> ExponentialBackoff {
         ExponentialBackoff::builder()
             .retry_bounds(self.min_interval, self.max_interval)
             .base(self.base)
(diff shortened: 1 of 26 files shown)


dependencies/pasqal_common/src/retry.rs
```diff
@@ -0,0 +1,316 @@
...
+// that they have been altered from the originals.
+
+//! Retry configuration shared by the Pasqal clients.
+//!
+//! Retries are split into two tiers:
+//!
+//! * ordinary transient failures (5xx, timeouts, connection errors), handled by
+//!   the stock [`RetryTransientMiddleware`] with an exponential backoff, and
+//! * rate-limit / locked responses (HTTP 429 and 423), handled by
+//!   `RetryAfterMiddleware`, which honors the server's `Retry-After` header
```
badtst: nit:
```suggestion
//!   [`RetryAfterMiddleware`], which honors the server's `Retry-After` header
```

dependencies/pasqal_common/src/retry.rs
```diff
@@ -0,0 +1,316 @@
...
+            }
+
+            // `should_retry` enforces the total-duration budget and provides the
+            // fallback backoff when the server sends no `Retry-After` header.
+            let RetryDecision::Retry { execute_after } =
+                self.policy.should_retry(start, n_past_retries)
+            else {
+                return Ok(response);
+            };
+            let now = SystemTime::now();
```
badtst: Not an expert here, but is timezone information handled ok here with regards to HTTP-date timezones that could be in headers ?

dependencies/pasqal_common/src/retry.rs
```diff
@@ -0,0 +1,316 @@
...
+            RetryStrategyExcept429,
+        ))
+        .with(RetryAfterMiddleware {
+            policy: config.rate_limit.build_policy(),
+            max_total_duration: config.rate_limit.max_total_duration,
+        })
+}
+
+#[cfg(test)]
+mod tests {
```
badtst, line 325: nit: move in `tests/retry.rs` ? Even if `parse_after_retry` is a "private" method ?

src/pasqal.rs
```diff
@@ -18,3 +18,26 @@ mod local;
...
+/// `yes`, or `on` (case-insensitive) disables retries; anything else (or unset)
+/// leaves the default retry policy in place.
+fn retries_disabled() -> bool {
+    let is_truthy = |v: String| {
+        matches!(
+            v.trim().to_ascii_lowercase().as_str(),
+            "1" | "true" | "yes" | "on"
+        )
+    };
+    env::var(RETRIES_DISABLED_ENV)
```
badtst: If you intend to set this parameter in `qrmi_config.json` note that this environment variable will be prefixed with the configuration's name i.e. `PASQAL_LOCAL_QRMI_PASQAL_RETRIES_DISABLED`. So I would try to adapt this function to match against those, otherwise it won't work

badtst (approved):
LGTM, I'd suggest adapting the way we match against the `QRMI_PASQAL_RETRIES_DISABLED` env variable for usage with the qrmi_config.json

dependencies/pasqal_common/src/retry.rs
sarjevane: The `Retry-After` header is defined [here](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3) as GMT so we do all the calculations with timezone-agnostic dates/durations, no problems here

dependencies/pasqal_common/src/retry.rs
sarjevane: That would require to make the method public, for a unit test IMO we can keep it here

src/pasqal.rs
sarjevane: Fixed, thanks for the catch! I broke it while refactoring and I missed it.

src/pasqal.rs
```diff
@@ -21,23 +21,31 @@ pub use self::local::PasqalLocal;
...
 
-/// Env var that, when set to a truthy value, disables HTTP retries for all
-/// Pasqal clients. This is a global toggle; it is not scoped per backend.
+/// Env var (unprefixed) that, when set to a truthy value, disables HTTP retries
+/// for a Pasqal client. QRMI namespaces backend configuration under a
+/// `<backend_name>_` prefix, so this is resolved as
+/// `<backend_name>_QRMI_PASQAL_RETRIES_DISABLED` (e.g.
+/// `PASQAL_LOCAL_QRMI_PASQAL_RETRIES_DISABLED`), falling back to the unprefixed
+/// name as a global override (e.g. for e2e tests).
 const RETRIES_DISABLED_ENV: &str = "QRMI_PASQAL_RETRIES_DISABLED";
```
badtst, line 28: An opinion on the naming @MatthieuMoreau0 @awennersteen ?

badtst (approved):
lgtm ! Waiting for feedback on the variable name

src/pasqal.rs
MatthieuMoreau0: An alternative suggestion: instead of activating/disabling retries with a boolean env variable, have you considered using this to define the maximum number of retries? It can be set to 0 to disable retries but can also be tweaked for finer controls over the retries

awennersteen: I like Matthieus suggestion.

dependencies/pasqal_common/src/retry.rs
```diff
@@ -0,0 +1,316 @@
...
+/// not exposed for per-call tuning by design — retries are toggled on/off at
+/// the client level, nothing more.
+#[derive(Debug, Clone)]
+struct RetryConfig {
+    /// Backoff for ordinary transient failures (5xx, timeouts, connection
+    /// errors). Excludes 429/423, which are handled by [`Self::rate_limit`].
+    transient: BackoffConfig,
+    /// Backoff for rate-limit (429) and locked (423) responses.
+    rate_limit: BackoffConfig,
+}
```
awennersteen, line 115: I assume that for auth errors it does not retry, but could you add that to the comment? Along with how that is ensured

dependencies/pasqal_common/src/retry.rs
```diff
@@ -0,0 +1,316 @@
...
+
+impl Default for RetryConfig {
+    /// Sensible production defaults: spend at most two minutes retrying
+    /// ordinary transient failures, and up to five minutes on rate limits.
+    fn default() -> Self {
```
awennersteen: the main issue I have here is that this will run as part of the scheduling logic by the system scheduler. 

I think first and foremost we need to ensure that retries are not attempted from within the scheduling logic (e..g inside spank plugins).

src/pasqal.rs
sarjevane: I had not considered the amount of retries, only the time budget, given the initial use-case - retry in case of transient errors to the Cloud.

IMHO a time budget (as a single value) is easier to reason with than an initial delay + exponential backoff + number of retries, so we could make the time-budget configurable (instead of simple on/off switch) - that would be my preference.

We can otherwise configure the number of retries, but that alone is only half of the story, we probably would need to make the initial delay configurable - but that feels like over-engineering to me, most users will not tweak these much.

Happy to discuss this further offline

dependencies/pasqal_common/src/retry.rs
sarjevane: It won't be part of the scheduling logic anymore, since only the python bindings now build the client with the retry middleware.

dependencies/pasqal_common/src/retry.rs
sarjevane: Under the hood we use https://docs.rs/reqwest-retry/0.7.0/reqwest_retry/fn.default_on_request_success.html , so all client errors are never retried (added a comment as well)

src/pasqal.rs
sarjevane: We now have a unified Middleware, which uses retries. If the number of retries is set to 0, then there is no retries (also in case of 429)

dependencies/pasqal_common/src/retry.rs
badtst: I'm still not 100% convinced that we should remove retries in the spank plugin (or other workload manager plugins) logic. We could argue that it's even the most sensitive part of the job execution workflow. 

If we fail on a transient error, the plugin fails to acquire resources and a user that maybe had to wait a certain amount of time to access the QPU is sent back at the end of the cluster's queue :/ Maybe we should seek feedback from the IBM team if they have had exeprience with retries in their `IBMQuantumSystem` implementation and the spank logic

sarjevane:
~~The only failure is a timeout on the schedmd/slurm website which is down at the moment - https://github.com/qiskit-community/qrmi/actions/runs/29724719180#summary-88297102759 - unrelated to the changes of this PR.~~

--------

Edit: schedmd's website is back up, re-running the failed test fixed it.

badtst:
When resuming work on this PR, we'll need to check wether the latest changes regarding the new error types had an impact on the handling of retries #212 (but looks like those were IBM-specific)

ohtanim:
> When resuming work on this PR, we'll need to check wether the latest changes regarding the new error types had an impact on the handling of retries #212 (but looks like those were IBM-specific)

IBM implementations stopped to use reqwesr-retry because it doesn’t return underlying status code.

PR #156, opened 2026-07-06
[PASQAL] feat: Support self-signed SSL certs

sarjevane:
## Description of Change

Swap out `rustls-tls` for `rustls-tls-native-roots` ([docs](https://docs.rs/reqwest/0.12.14/reqwest/#optional-features)) to support custom CAs and self-signed certificates.
Update documentation describing what to change to support self-signed certs.

## Checklist ✅

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2905,6 +2905,7 @@ dependencies = [
  "pin-project-lite",
  "quinn",
  "rustls 0.23.40",
+ "rustls-native-certs",
  "rustls-pki-types",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


MatthieuMoreau0 (approved):

awennersteen (approved):
lgtm. is this relevant to also enable for pasqal cloud client?

sarjevane:
> lgtm. is this relevant to also enable for pasqal cloud client?

The use-case here was for the introduction of a reverse proxy with TLS in front of Warden with self-signed certificates for a deployment on a QAN in an HPC center.

The cloud will always have its own verified certificates (through Let's Encrypt or similar mechanism) with a valid chain, so we would not need to relax the requirements to allow for custom/OS CAs like we did here.

The only use-case that we would have a cloud with self-signed certificates would be in a context of a TLS reverse proxy in front of a "cloud" running locally (e.g. a mock for e2e-tests), but I doubt we will ever have that need - we would just run over HTTP without reverse proxy

TL;DR:
IMHO not relevant for cloud

PR #162, opened 2026-07-13
Use a callback to expose logs to scheduler logs

awennersteen:
## Description of Change

The goal here was to make the QRMI logs called from e.g. the spank plugins appear in their logs.

I propose to solve this via a callback in the Rust core. 

Another option considered was to write to  a QRMI log file. 
However, since QRMI runs under both the Slurm daemon and end-user processes. A shared /var/log/qrmi.log would require careful permissions, ownership, etc, and  could expose user-job messages to unintended readers. Scheduler logs already provide the appropriate system-level destination and access controls while keeping job-related diagnostics available in the relevant job or daemon logs.

Note: I have some issues getting the spank plugin side up right now, let's not merge this before both sides are up and we can all test it properly

Not tested after cutting this out from other WIP code, but this should work.
https://github.com/qiskit-community/spank-plugins/pull/196

After this PR the slurmd logs look like the below, with the qrmi logs correctly :

```
[2026-07-14T08:14:25.034] [52.batch] debug2: Before call to spank_fini()
[2026-07-14T08:14:25.034] [52.batch] debug:  spank_qrmi(341, 0): -> slurm_spank_exit argc=2 remote=1
[2026-07-14T08:14:25.034] [52.batch] debug:  spank_qrmi: releasing name(PASQAL_LOCAL), type(4), token(81ae2c81-7ceb-4248-8cf0-0c90ffbd1f10)
[2026-07-14T08:14:25.035] [52.batch] debug:  spank_qrmi QRMI reqwest::connect: starting new connection: http://c1:4207/
[2026-07-14T08:14:25.040] [52.batch] debug:  spank_qrmi QRMI pasqal_local_api::client: {"id":"81ae2c81-7ceb-4248-8cf0-0c90ffbd1f10","user_id":"1000","created_at":"2026-07-14T08:14:24.013553","revoked_at":"2026-07-14T08:14:25.037658Z","qpu_slots":1}
[2026-07-14T08:14:25.040] [52.batch] debug:  spank_qrmi(341,0): <- slurm_spank_exit
[2026-07-14T08:14:25.040] [52.batch] debug2: spank: spank_qrmi.so: exit = 0
[2026-07-14T08:14:25.040] [52.batch] debug2: After call to spank_fini()
[2026-07-14T08:14:25.040] [52.batch] job 52 completed with slurm_rc = 0, job_rc = 0
[2026-07-14T08:14:25.041] [52.batch] debug2:   false, shutdown
[2026-07-14T08:14:25.041] [52.batch] debug:  Message thread exited
[2026-07-14T08:14:25.041] [52.batch] done with step
[2026-07-14T08:14:25.042] debug2: Start processing RPC: REQUEST_TERMINATE_JOB
[2026-07-14T08:14:25.042] debug2: Processing RPC: REQUEST_TERMINATE_JOB
[2026-07-14T08:14:25.042] debug:  _rpc_terminate_job: uid = 990 JobId=52
[2026-07-14T08:14:25.042] debug:  credential for job 52 revoked
[2026-07-14T08:14:25.042] debug2: No steps in jobid 52 to send signal 18
[2026-07-14T08:14:25.042] debug2: No steps in jobid 52 to send signal 15
[2026-07-14T08:14:25.042] debug2: set revoke expiration for jobid 52 to 1784016985 UTS
[2026-07-14T08:14:25.042] debug2: Finish processing RPC: REQUEST_TERMINATE_JOB
```

Here we see a potential token leak -  will fix that in a separate MR, it's in existing code.

## Checklist ✅

- [ x ] Have you included a description of this change?
- [ x ] Have you updated the relevant documentation to reflect this change?
- [ x ] Have you made sure CI is passing before requesting a review?

## Ticket
- [ ] Fixes #
- [ ] Is Part of #


OkuyanBoga (approved):

src/common.rs
```diff
@@ -9,13 +9,60 @@
...
-use std::sync::Once;
+use std::ffi::CString;
+use std::io::Write;
+use std::os::raw::c_char;
+use std::sync::{Mutex, Once};
 
 static INIT: Once = Once::new();
+static LOG_CALLBACK: Mutex<QrmiLogCallback> = Mutex::new(None);
+
+pub type QrmiLogCallback = Option<
```
cclaudio, line 20: I'd suggest renaming it to something that reflects its unsafe behavior, e.g. `QrmiFfiLogCallback`

src/common.rs
```diff
@@ -9,13 +9,60 @@
...
 // copyright notice, and modified files need to carry a notice indicating
 // that they have been altered from the originals.
-use std::sync::Once;
+use std::ffi::CString;
+use std::io::Write;
+use std::os::raw::c_char;
+use std::sync::{Mutex, Once};
 
 static INIT: Once = Once::new();
+static LOG_CALLBACK: Mutex<QrmiLogCallback> = Mutex::new(None);
```
cclaudio: Maybe FFI_LOG_CALLBACK ?

src/common.rs
```diff
@@ -9,13 +9,60 @@
...
+
+pub type QrmiLogCallback = Option<
+    unsafe extern "C" fn(level: *const c_char, target: *const c_char, message: *const c_char),
+>;
+
+pub(crate) fn set_log_callback(callback: QrmiLogCallback) {
+    if let Ok(mut current) = LOG_CALLBACK.lock() {
+        *current = callback;
```
cclaudio: Do we need to ensure that it is set only once?

src/common.rs
ohtanim: I disagree. All C FFI operations are unsafe. If we apply this naming convention here, we should apply it consistently and include FFI in the names of all such functions.

src/common.rs
ohtanim: I don't think so.

src/common.rs
```diff
@@ -9,13 +9,60 @@
...
+    }
+}
+
+fn callback_log(record: &log::Record<'_>) -> bool {
```
ohtanim: The function `callback_log(record)` passes a record to the registered callback (returning true/false for whether it succeeded), but going by the name alone, it could also be read the opposite way (as if it's "calling back" the log, or logging a callback?). Names like `dispatch_to_callback` or `try_log_via_callback `might convey the intent more clearly. This is just a matter of preference.

(I prefer to use `dispatch_to_callback`)

src/cext.rs
```diff
@@ -179,6 +181,15 @@ fn _set_last_error(msg: String) {
     });
 }
 
+/// Registers a QRMI log callback for C hosts.
+///
+/// Pass NULL to clear the current callback.
+#[no_mangle]
+pub unsafe extern "C" fn qrmi_log_callback_set(callback: QrmiLogCallback) {
```
ohtanim: This function should return a ReturnCode, like the other C FFI functions, because it can fall.

```
#[no_mangle]
pub unsafe extern "C" fn qrmi_log_callback_set(callback: QrmiLogCallback) -> ReturnCode {
    let result = crate::common::set_log_callback(callback);
    crate::common::initialize();
    match result {
        Ok(()) => ReturnCode::Success,
        Err(()) => ReturnCode::Error,
    }
}
```

src/cext.rs
```diff
@@ -179,6 +181,15 @@ fn _set_last_error(msg: String) {
     });
 }
 
+/// Registers a QRMI log callback for C hosts.
```
ohtanim, line 185: Please enhance documentation. Since this is unsafe FF, at very leasy, "Safety" section should be added.


```
/// @ingroup Qrmi
/// Registers a QRMI log callback for C hosts.
///
/// Once registered, all log records produced by this library (subject to
/// the active `RUST_LOG` filter, default level `warn`) are routed to
/// `callback` instead of stderr.
///
/// Pass NULL to clear the current callback and revert to the default
/// stderr writer.
///
/// # Safety
///
/// * If `callback` is non-NULL, it must remain a valid, callable function
///   pointer for as long as it might still be invoked. Because a
///   currently-executing log call may have already captured the previous
///   callback pointer, do not unload code backing a callback immediately
///   after replacing or clearing it — a small number of in-flight calls
///   may still land on the old pointer.
///
/// # Example
///
/// @code
///   void my_log_cb(const char *level, const char *target, const char *message) {
///     fprintf(stderr, "[%s] %s: %s\n", level, target, message);
///   }
///   QrmiReturnCode rc = qrmi_log_callback_set(my_log_cb);
/// @endcode
///
/// @param (callback) [in] Callback function, or NULL to clear.
/// @return @ref QrmiReturnCode::QRMI_RETURN_CODE_SUCCESS if succeeded.
/// @version 0.6.0
```

src/common.rs
```diff
@@ -9,13 +9,60 @@
...
+    }
+}
 
 /// Called once before using the API library to initialize static resources(logger etc.) in underlying layers. If called more than once, the second and subsequent calls are ignored.
 pub(crate) fn initialize() {
     INIT.call_once(|| {
-        env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("warn")).init();
+        env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("warn"))
+            .format(|buf, record| {
+                if callback_log(record) {
```
ohtanim: Please consider to use `dispatch_to_callback` or `try_log_via_callback`.

src/common.rs
```diff
@@ -9,13 +9,60 @@
...
+}
+
+fn callback_log(record: &log::Record<'_>) -> bool {
+    let callback = LOG_CALLBACK.lock().ok().and_then(|current| *current);
+    let Some(callback) = callback else {
+        return false;
+    };
+    let level = CString::new(record.level().as_str()).ok();
+    let target = CString::new(record.target()).ok();
+    let message = CString::new(record.args().to_string()).ok();
```
ohtanim: If any of `level`/`target`/`message` contains a `NUL` character, `callback_log` returns `false`. This is because Rust's `CString::new` returns an error (`NulError`) if the string contains a `NUL` character (\0) anywhere other than at the very end. In that case, that particular record falls back to the normal `env_logger` format (timestamped text). Normally, when a callback is registered, everything should be funneled through it — but in this one case, the record gets output through a different path (e.g. `stderr`), which is an inconsistency.
How would you like to handle strings that contain NUL characters? I'd prefer to strip out the `NUL` characters and write the result.

src/common.rs
ohtanim: The mutex gets locked on every log call (across all threads, all log levels). The lock hold time itself is extremely short, so this normally isn't an issue, but under high-frequency logging (e.g. a job-status polling loop) with many threads logging concurrently, this could become a bottleneck. Since the callback is replaced infrequently and is essentially read-mostly, something like RwLock would be better suited here. That said, if performance isn't a concern at this scale, the current implementation should be fine in practice.

`static LOG_CALLBACK: RwLock<QrmiLogCallback> = RwLock::new(None);`

src/common.rs
ohtanim:
```
fn sanitized_cstring(s: &str) -> CString {
    if s.as_bytes().contains(&0) {
        let cleaned: Vec<u8> = s.bytes().filter(|&b| b != 0).collect();
        CString::new(cleaned).unwrap_or_default()
    } else {
        CString::new(s).unwrap_or_default()
    }
}
```

ohtanim (changes requested):
Thanks for your suggestion on enhancement.

src/common.rs
awennersteen: good suggestion, implemented

src/cext.rs
awennersteen: Implemented. good catch

src/cext.rs
awennersteen: thanks!

src/common.rs
awennersteen: implemented

src/common.rs
awennersteen: yes, good idea.  This I had not thought through. 

Implemented

src/common.rs
awennersteen: It is an easy thing to move to RwLock, so I did it. After reading about it, it should indeed be at least not worse than a mutex

Implemented

src/common.rs
awennersteen: Good concern.
  The logger itself is initialized only once via Once; the callback registration is separate.

  The callback setter supports:

  - Registering a callback.
  - Replacing an existing callback.
  - Clearing it by passing NULL.

The callback pointer update is protected by RwLock. Restricting registration to one call would conflict with it's behavior and make callback lifecycle management less useful.

src/common.rs
awennersteen: Let's discuss above

src/common.rs
awennersteen: I personally don't see the benefit of including FFI in the naming convention, but I have no argument against it either.  But as Othani says, all QRMI functions will have the same level of unsafeness, right?

ohtanim: Yes. You can find "unsafe" tag in the source code, which means we have the same level of unsafeness.

cclaudio: As a developer, I saw the callback variable carrying an unsafe function then I started thinking how we could reflect the unsafe behavior to whoever touches that variable. That's why I suggested renaming the variable as possible solution.

That's fine no need to rename it.  The `Safety` comment in the consumer addresses my concern. Thanks

ohtanim (approved):
LGTM

Issue #168, opened 2026-07-21
[EPIC] API versioning


cclaudio:
### What problem are you trying to solve?

QRMI provides an API that can be used for things like task execution, accounting, resource acquisition and resource discovery.

Currently, we are using the QRMI crate version to represent the QRMI API version. However, it is also used to track changes in other parts of the project such as API implementation and internal code. So, at the end of the day, a QRMI version bump does not mean that the QRMI API has changed. With that, it is difficult determine what is the current API version and what is included in it. 

Ideally, we should
- Isolate the QRMI API code so that we can easily determine what is included it
- Use a separate version counter for it so that we can keep track of changes

### Why should we solve this problem?
 
- As a user/developer/admin/researcher, I'd like to know the current API version and its specification so that I can evaluate/study it. That includes checking what changed from a previous version.
- As an admin, I'd like to know if a QRMI version changed the API so that I can debug and detect compatibility issues.
- As a user, I'd like to have a stable API so that I don't need to upgrade my deployment often.
- As an API mantainer, I'd like to easily determine what is included in the current API version so that I can keep track of API changes and also have PRs automatically assigned to me if they are changing the API.
- As a QRMI caller (e.g. slurm plugin), I'd like to query the current API version at runtime so that I can detect compatibility issues early. E.g. if the API version is too old or too recent, the slurm plugin should fail to load and print error saying that the API version is not supported.

### Describe the solution you’d like

We could isolate all the QRMI API definitions in a separate rust crate (e.g. qrmi-core-api) and then use crate version to represent the API version

Since we have bindings for other languages, we may need to have separate crates for them too. If we need to fix somethingthat is specific for the binding, we would need to bump only that binding crate without bumping the other bindings or the core API defined in rust.

Work items:
- [ ] #169
- [ ] Maintain bindings in separate crates (e.g. qrmi-cext and qrmi-pyext) and use their version to keep track of binding changes.
- [ ] Add an API function like `api_version()` to get the QRMI API version

### Describe any alternatives you’ve considered

I experimented with maintaining the API version using just a simple variable (details in the branch [here](https://github.com/cclaudio/qrmi/tree/api-version-variable)). However, I found some issues with that.
[...]

PR #169, opened 2026-07-21
API versioning by isolating the API source code in a new crate

cclaudio:
## Description of Change

- Create a folder named `crates` to be the home of new crates. A flat structure is known to be more beneficial than hierarchical.
- Move API files to qrmi-core-api.
- Use CI to ensure that API changes are not released without bumping the qrmi-core-api version.
- Add qrmi-core-api to `workspace.dependencies` to ensure that workspace members use the same qrmi-core-api version

## Checklist ✅

## Ticket
- [ ] Fixes #
- [x] Is Part of #168


ohtanim (commented):
@cclaudio Have you read my comment in your git issue?
https://github.com/qiskit-community/qrmi/issues/168#issuecomment-5038182822

I would like to understand why you're proposing creating separate crates instead of creating git branches for tracking API changes. Git branches handle this well. The common pattern is to create a maintenance branch per major version.

I'm not sure which of the git issues you wrote this crate is meant to solve. Additionally, suppose that going forward there will be multiple major versions of the API. Security fixes and patches will need to be issued for each of them. How does this crate separation solve that use case?

cclaudio:
> @cclaudio Have you read my comment in your git issue? [#168 (comment)](https://github.com/qiskit-community/qrmi/issues/168#issuecomment-5038182822)

Yes, I replied to your comment there.

> I would like to understand why you're proposing creating separate crates instead of creating git branches for tracking API changes. Git branches handle this well. The common pattern is to create a maintenance branch per major version.
> 
> I'm not sure which of the git issues you wrote this crate is meant to solve. Additionally, suppose that going forward there will be multiple major versions of the API. Security fixes and patches will need to be issued for each of them. How does this crate separation solve that use case?

The QRMI version can be used to track changes in general, including changes to API implementation and internal code. By isolating the API in the `qrmi-core-api` crate, we can use the `qrmi-core-api` version to track changes specific to the API.

ohtanim:
@awennersteen @MatthieuMoreau0 @OkuyanBoga @yoonho @machjn Hi Reviewers,

We're discussing this on the git issue #168 — please share your thoughts there if you have any.

I've already posted my opinion, but to summarize: 1) the benefits gained from splitting into crates are small, or fall short of what's expected; 2) even with the crate split, managing the API version still requires manual steps, so there's no significant difference in effort; 3) rather than the proposed approach, splitting by branches — as is common practice in other open source projects — has more advantages.

cclaudio:
Closing this PR, it is not needed. More details in https://github.com/qiskit-community/qrmi/issues/168#issuecomment-5206984015

PR #188, opened 2026-08-13
Bridge Rust log records to Python's logging module (and fix a GIL deadlock in blocking calls)

ohtanim:
### Summary

QRMI's Python extension (`pyext.rs`) registers a callback with the shared `log`-crate sink (`common.rs`) so that `log::error!()` and friends — previously visible only on the C/CLI path — now reach Python's `logging` module by default, and can be redirected to a user-supplied Python callable via a new `qrmi.set_log_callback()` function. Along the way this also fixes a real deadlock that the naive version of this change introduced, and refactors `common.rs` so it no longer forces a C ABI shape onto its pure-Rust consumer.

No public C API behavior changes. `qrmi_log_callback_set()`'s contract (NULL semantics, string arguments, in-flight-call caveat) is unchanged.

### Motivation

Previously, `log::error!()` records produced by this crate (e.g. inside `target()`, `is_accessible()`) were only visible when QRMI was used from the CLI/examples or from C, via `env_logger`'s stderr output. When used as a Python library — the primary way most users consume QRMI — no logging backend was ever registered, so these records were silently dropped. This made real failures (backend errors, malformed responses) invisible unless they also happened to surface as a raised exception at the same call site.

#### 1. `common.rs`: made the shared log sink Rust-native

`common.rs` previously defined the single registration point for "where do log records go" (`LOG_CALLBACK` / `QrmiLogCallback`) as a C ABI type (`unsafe extern "C" fn(*const c_char, ...)`), because that was the only
shape `cext.rs`'s C-facing API could hand it. `pyext.rs` had no need for that shape at all, but had to conform to it anyway, round-tripping every record through `CString`/`CStr` for no reason.

`common.rs` now exposes a plain Rust type instead:

```rust
pub(crate) type LogSink = Arc<dyn Fn(log::Level, &str, &str) + Send + Sync>;
pub(crate) fn set_log_sink(sink: Option<LogSink>) -> Result<(), ()>;
```

`cext.rs` and `pyext.rs` each register their own adapter against this one Rust-native type; the C ABI shape now lives only in `cext.rs`, where it's actually needed.

`LogSink` is `Arc`, not `Box`: `dispatch_to_sink` clones the handle and releases `LOG_SINK`'s `RwLock` *before* invoking it, rather than holding the lock for the call's duration. This matters once a sink can run arbitrary, possibly slow, possibly re-entrant code (a user's Python callback in particular) — `std::sync::RwLock` does not guarantee safe
recursive read-locking, so holding it across such a call would risk a deadlock or worse. (The previous C-function-pointer design didn't have this problem, because a `Copy` function pointer could be read out of the lock and the lock dropped immediately; a boxed closure can't be copied out that way.)

#### 2. `cext.rs`: `QrmiLogCallback` moved here, C string handling stays here

`QrmiLogCallback` (the C function pointer type) and the `CString` sanitization helper (`sanitized_cstring`, stripping embedded NULs before building a `CString`) now live in `cext.rs` instead of `common.rs`. `qrmi_log_callback_set()` builds a small adapter closure around the caller's C function pointer, doing the `CString` conversion only when a
record actually needs to reach it, and registers that adapter as a `LogSink`. The function's public behavior and safety contract are unchanged.

#### 3. `pyext.rs`: default logging bridge + user-overridable callback

- A default sink (`python_log_sink`) is registered when the `_core` module is imported. It forwards each record to
  `logging.getLogger(target).log(level, message)`, so standard Python logging configuration (`logging.basicConfig()`, per-logger levels, handlers) governs QRMI's Rust-side log output the same way it governs any other Python log record. No action needed from users who just want
  logging to work.
- A new function, `qrmi.set_log_callback(callback)`, lets a user register their own `callback(level: str, target: str, message: str)` in place of the default, mirroring the C API's `qrmi_log_callback_set()`. Passing `None` clears it, reverting to plain stderr (the same fallback `NULL` triggers on the C side) rather than back to the `logging`-forwarding default — once a caller takes over, where records go is entirely their callback's responsibility. If the callback raises, the exception is printed (as an unhandled exception would be) and discarded; it cannot propagate back into the Rust call site that produced the log record.
- Rust's `RUST_LOG` filter (default `warn`) still runs before any sink is invoked, Python or C, and currently cannot be changed from the Python side; only records that pass it reach `logging` (or a user callback) at all. This is a known, separate limitation, not addressed by this PR.
- Python code inside this package that calls `logging.getLogger("qrmi")` directly (e.g. `python/qrmi/primitives/service.py`) is unaffected by any of this — those records never go through the Rust `log` crate or `LogSink` in the first place, so `set_log_callback()` has no effect on them.

#### 4. Deadlock fix: release the GIL around every blocking call

While testing the above with `RUST_LOG=trace`, we hit a real deadlock: `PyQuantumResource`'s methods call `self.rt.block_on(...)` on a multi-threaded tokio `Runtime` while still holding the GIL (standard for a
`#[pymethods]` call). At `trace` level, `reqwest`/`hyper` log frequently enough that some record is virtually guaranteed to be emitted from a tokio worker thread. That worker thread then needs the GIL to reach `python_log_sink`/a user callback — but the GIL is held by the main thread, which is itself blocked inside `block_on` waiting on that same worker thread. Both sides wait forever.

Fixed by wrapping every `self.rt.block_on(...)` call (14 call sites across `PyQuantumResource` and `PyResourceProvider`) in `py.detach(|| ...)`,  releasing the GIL for the duration of the blocking wait — the documented
pattern for this exact situation (see `Python::detach`'s own docs, which name "awaiting a future" as a case to release the GIL for). This was verifiable only by actually building and running with `RUST_LOG=trace`; it would not have been caught by `warn`-level testing, since error/warn records are rare enough that the race window is unlikely to be hit.

### Testing

- Built and run manually against live IBM Qiskit Runtime Service backends   (`examples/qiskit_primitives/ibm/sampler.py`), confirming:
  - Default behavior: `logging.basicConfig(level=logging.ERROR, ...)` in the calling script now shows QRMI's Rust-side `error!()` records with the expected target (`qrmi::ibm::qiskit_runtime_service`) and message.
  - `RUST_LOG=trace` no longer hangs (previously deadlocked before the `py.detach()` fix).
  - A custom callback registered via `qrmi.set_log_callback()` receives Rust-side records instead of the default `logging` forwarding, and Python-native `qrmi` package logging (`primitives/service.py` etc.) is unaffected either way, as expected.
  
- Test with [Spank plugins (C API)  PR](https://github.com/qiskit-community/spank-plugins/pull/196)

### Backward compatibility

- **C API**: no change. `qrmi_log_callback_set()` keeps the same signature, NULL semantics, and in-flight-call caveat.
- **Python API**: additive only. Existing code that does nothing keeps getting the same behavior as before, except that `error!()`-level records are now actually visible (previously silent) via standard `logging` configuration. `qrmi.set_log_callback()` is a new, optional function.
- **Rust internals**: `common::QrmiLogCallback` and `common::set_log_callback()` no longer exist; both moved to `cext.rs` (renamed `set_log_sink`/`LogSink` at the `common.rs` level). This only affects code inside this crate — `cext.rs` and `pyext.rs` were the only callers, and both are updated in this PR.

### Limitations

- `RUST_LOG`'s level filter is still separate from Python's per-logger levels; a Python user cannot currently raise the Rust-side filter above `warn` without setting the environment variable before the process starts. Worth a follow-up if `DEBUG`/`TRACE` visibility from pure Python is wanted.

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/src/pyext.rs
+++ b/src/pyext.rs
@@ -40,7 +40,31 @@ pub enum ResourceType {
 #[pyo3(name = "QuantumResource")]
 pub struct PyQuantumResource {
     qrmi: Box<dyn QuantumResource + Send + Sync>,
-    rt: Runtime,
+    // `ManuallyDrop`, not a plain `Runtime`, so `Drop` below can take
+    // ownership and call `shutdown_background()` instead of letting the
+    // field's own destructor run. Existing `self.rt.block_on(...)` call
+    // sites are unaffected: `ManuallyDrop<T>` derefs to `T` transparently.
+    rt: std::mem::ManuallyDrop<Runtime>,
+}
+
+impl Drop for PyQuantumResource {
+    fn drop(&mut self) {
+        // `Runtime`'s own `Drop` blocks the current thread indefinitely
+        // until every spawned task finishes. That is a problem here
+        // specifically: Python drops objects while holding the GIL, and
+        // if a worker thread needs the GIL to log something (e.g. an
+        // in-flight HTTP request logged at `RUST_LOG=trace`) before it can
+        // finish, that worker waits for the GIL forever while this thread
+        // waits for that worker forever. `shutdown_background()` discards
+        // the runtime without waiting for anything, which avoids the
+        // deadlock entirely (in exchange for not waiting for in-flight
+        // background work to finish cleanly on drop).
+        //
+        // SAFETY: `self.rt` is only ever taken here, in `Drop::drop`,
+        // which runs at most once per instance.
+        let rt = unsafe { std::mem::ManuallyDrop::take(&mut self.rt) };
+        rt.shutdown_background();
+    }
 }
 
 impl PyQuantumResource {
@@ -48,7 +72,9 @@ impl PyQuantumResource {
     pub(crate) fn from_inner(qrmi: Box<dyn QuantumResource + Send + Sync>) -> Self {
         Self {
             qrmi,
-            rt: Runtime::new().unwrap(),
+            rt: std::mem::ManuallyDrop::new(
+                Runtime::new().expect("Failed to create a new tokio runtime."),
+            ),
         }
     }
 }
@@ -102,7 +128,9 @@ impl PyQuantumResource {
 
         Ok(Self {
             qrmi,
-            rt: Runtime::new().unwrap(),
+            rt: std::mem::ManuallyDrop::new(
+                Runtime::new().expect("Failed to create a new tokio runtime."),
+            ),
         })
     }
 
@@ -313,7 +341,20 @@ impl PyResourceDef {
 #[pyo3(name = "ResourceProvider")]
 pub struct PyResourceProvider {
     inner: Box<dyn crate::ResourceProvider>,
-    rt: Runtime,
+    // See `PyQuantumResource`'s `rt` field and `Drop` impl for why this is
+    // `ManuallyDrop` rather than a plain `Runtime`.
+    rt: std::mem::ManuallyDrop<Runtime>,
+}
+
+impl Drop for PyResourceProvider {
+    fn drop(&mut self) {
+        // See `PyQuantumResource`'s `Drop` impl for why.
+        //
+        // SAFETY: `self.rt` is only ever taken here, in `Drop::drop`,
+        // which runs at most once per instance.
+        let rt = unsafe { std::mem::ManuallyDrop::take(&mut self.rt) };
+        rt.shutdown_background();
+    }
 }
 
 #[gen_stub_pymethods]
@@ -349,7 +390,9 @@ impl PyResourceProvider {
         };
         Ok(Self {
             inner,
-            rt: Runtime::new().unwrap(),
+            rt: std::mem::ManuallyDrop::new(
+                Runtime::new().expect("Failed to create a new tokio runtime."),
+            ),
         })
     }
 
@@ -489,6 +532,15 @@ impl PyConfig {
 /// particular sink happens to call into Python; that's this module's
 /// business alone. Compare `cext::qrmi_log_callback_set`, which adapts a
 /// C function pointer into the same `LogSink` shape at its own boundary.
+///
+/// Uses `Python::try_attach`, not `Python::attach`: a log record can be
+/// emitted from a `__del__` running during CPython interpreter
+/// finalization (`Py_FinalizeEx`) -- attempting to (re-)acquire the GIL
+/// in that window is a known hazard (documented on `Python::attach`
+/// itself, and the subject of e.g. PyO3#5317: repeatedly attaching during
+/// finalization has hung or segfaulted). `try_attach` returns `None`
+/// instead in that case; we simply drop the record rather than risk
+/// hanging the interpreter shutdown to deliver a log line.
 fn python_log_sink(level: log::Level, target: &str, message: &str) {
     // Python's `logging` module level numbers (see Python's `logging`
     // module docs / `Lib/logging/__init__.py`: CRITICAL=50, ERROR=40,
@@ -504,7 +556,7 @@ fn python_log_sink(level: log::Level, target: &str, message: &str) {
         log::Level::Debug | log::Level::Trace => 10,
     };
 
-    Python::attach(|py| {
+    Python::try_attach(|py| {
         let Ok(logging) = py.import("logging") else {
             return;
         };
@@ -544,6 +596,10 @@ fn python_log_sink(level: log::Level, target: &str, message: &str) {
 /// - If `callback` raises, the exception is printed (as an unhandled
 ///   exception would be) and otherwise discarded; it cannot propagate
 ///   back into the Rust code that logged in the first place.
+/// - Like `python_log_sink`, this uses `Python::try_attach`: if a record
+///   is emitted while the interpreter is finalizing, `callback` is simply
+///   not called for it rather than risking the same hang/segfault hazard
+///   `Python::attach` has in that window.
 #[gen_stub_pyfunction]
 #[pyfunction]
 fn set_log_callback(callback: Option<Py<PyAny>>) -> PyResult<()> {
@@ -551,7 +607,7 @@ fn set_log_callback(callback: Option<Py<PyAny>>) -> PyResult<()> {
     let sink: Option<crate::common::LogSink> = callback.map(|callback| {
         let sink: crate::common::LogSink =
             std::sync::Arc::new(move |level: log::Level, target: &str, message: &str| {
-                Python::attach(|py| {
+                Python::try_attach(|py| {
                     if let Err(err) = callback.call1(py, (level.as_str(), target, message)) {
                         err.print(py);
                     }
--- a/src/pyext.rs
+++ b/src/pyext.rs
@@ -106,27 +106,27 @@ impl PyQuantumResource {
         })
     }
 
-    fn is_accessible(&mut self) -> PyResult<bool> {
+    fn is_accessible(&mut self, py: Python<'_>) -> PyResult<bool> {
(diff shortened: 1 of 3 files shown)


badtst:
Thanks Ohtani san, this is an extremely useful feature. I managed to test it locally and it will be a critical asset to help debug qrmi deployments.

badtst:
I have noticed that when setting the `reqwest` log level to `trace` and running the [`pulser_qrmi.py`](https://github.com/qiskit-community/qrmi/blob/main/examples/pulser/pasqal/pulser_qrmi.py) example that we had munge token leakage through the headers logged by `reqwest` [here](https://github.com/qiskit-community/qrmi/blob/main/dependencies/pasqal_local_client/src/client.rs#L92).

I tried setting the munge `HeadeValue`'s sensitivity level with [`set_sensitivity`](https://docs.rs/reqwest/latest/reqwest/header/struct.HeaderValue.html#method.is_sensitive) similarly to the way it is done in [`quantum_system_client`](https://github.com/qiskit-community/qrmi/blob/main/dependencies/quantum_system_client/src/client.rs#L515) but the token still appears in the `rewest` logs passed though the python sink.

 Could you confirm that you also have this token leakage with the `quantum_system_client`'s logs in python or an idea of where it could come from ?

Edit: It seems that setting the `reqwest` logs to trace does not take into account the `set_sensitivity` parameter. I think the next steps following this PR, since it will ease user's access to `trace/debug` logs from the qrmi would be to either:
- Remove the `connection_verbose` from the reqwest builder
- Sanitize the logs
- set the logging options/configuration that could leak secrets (such as the `set_sensitivity` method of the reqwest builder) only for `dev` qrmi builds and remove them from the `release` builds of the QRMI

Lmk what you think

ohtanim:
> I have noticed that when setting the `reqwest` log level to `trace` and running the [`pulser_qrmi.py`](https://github.com/qiskit-community/qrmi/blob/main/examples/pulser/pasqal/pulser_qrmi.py) example that we had munge token leakage through the headers logged by `reqwest` [here](https://github.com/qiskit-community/qrmi/blob/main/dependencies/pasqal_local_client/src/client.rs#L92).
> 
> I tried setting the munge `HeadeValue`'s sensitivity level with [`set_sensitivity`](https://docs.rs/reqwest/latest/reqwest/header/struct.HeaderValue.html#method.is_sensitive) similarly to the way it is done in [`quantum_system_client`](https://github.com/qiskit-community/qrmi/blob/main/dependencies/quantum_system_client/src/client.rs#L515) but the token still appears in the `rewest` logs passed though the python sink.
> 
> Could you confirm that you also have this token leakage with the `quantum_system_client`'s logs in python or an idea of where it could come from ?
> 
> Edit: It seems that setting the `reqwest` logs to trace does not take into account the `set_sensitivity` parameter. I think the next steps following this PR, since it will ease user's access to `trace/debug` logs from the qrmi would be to either:
> 
> * Remove the `connection_verbose` from the reqwest builder
> * Sanitize the logs
> * set the logging options/configuration that could leak secrets (such as the `set_sensitivity` method of the reqwest builder) only for `dev` qrmi builds and remove them from the `release` builds of the QRMI
> 
> Lmk what you think

@badtst Thanks for raising this. My understanding is the same as yours. In other words, `set_sensitive()` and `connection_verbose() `are completely unrelated. `set_sensitive()` only affects how a `HeaderValue` is displayed when logged via `debug!()`, but it does not affect the verbose connection logs. 

Simply removing `connection_verbose(true)` will prevent the raw connection read/write data from being included in the verbose logs. However, that also makes it harder to understand the interactions with the server when troubleshooting. Therefore, it might make sense to enable `connection_verbose(true)` only for debug builds and disable it for release builds.

For example, instead of:
```
reqwest_client_builder =
    reqwest_client_builder.connection_verbose(true);
```
(Edit) you could use:
```
#[cfg(debug_assertions)]
{
reqwest_client_builder =
    reqwest_client_builder.connection_verbose(true);
} 
```
This way, verbose connection logging is enabled only in debug builds. What do you think ?

badtst:
I think that's the best solution in this case, and it would set a good precedent when it comes to logging and information leakage prevention.

ohtanim:
> I think that's the best solution in this case, and it would set a good precedent when it comes to logging and information leakage prevention.

OK. Then, let's fix in separate PR. Would you be interested in this PR creation ?

badtst:
Yes no problem. I can handle this in a separate PR next week

ohtanim:
> Yes no problem. I can handle this in a separate PR next week

Thanks. I can help to prepare IBM side fix and test them. Please involve me when you create a PR.

ohtanim:
> Yes no problem. I can handle this in a separate PR next week

Opened #189 for this fix.

ohtanim:
Added @badtst to reviewer. 
@awennersteen @OkuyanBoga If you have a chance, I'd appreciate your feedback on the content.

src/pyext.rs
```diff
@@ -451,9 +509,123 @@ impl PyConfig {
...
+        sink
+    });
+    crate::common::set_log_sink(sink)
+        .map_err(|()| pyo3::exceptions::PyRuntimeError::new_err("Failed to set log callback"))
+}
+
 /// A Python module implemented in Rust.
 #[pymodule(name = "_core")]
 fn qrmi(m: &Bound<'_, PyModule>) -> PyResult<()> {
+    crate::common::initialize();
```
awennersteen, line 625: with this, the rust module will be initializsed when importing the Python module, right? are there any drawbacks to that?

awennersteen (approved):
lgtm with a superficial review. Only wanting to make sure that there are no drawbacks to the new position of the `inititialize`.

src/pyext.rs
ohtanim: @awennersteen Thanks for the question. Short answer: yes, moving common::initialize() to run at `import qrmi` time can change behavior compared to before. While digging into this, we found a more significant issue than we initially realized, which we're also reporting here.

Background: `log::set_logger()` (called internally by `env_logger`) only allows the first caller in the whole process to succeed. If some other Python module in the same process also bundles Rust code that sets up `env_logger` (or the log crate) the same way we do, whichever one registers first wins.

This race isn't new in this PR -- it already existed. Previously it was triggered the first time any QRMI Python method was called; this change simply moves that trigger earlier, to `import qrmi` time.

A more significant issue I found along the way: Good catch. Thanks for comments. we initially assumed this was a "rarely causes a behavior difference" kind of risk. In fact, `common.rs`'s use of `env_logger::Builder::init()` panics if another logger(outside of our crates) has already been registered. Worse, because `initialize()` is guarded by a `std::sync::Once`, a panic inside that initialization closure poisons the Once -- meaning every subsequent call to `common::initialize()` (which runs at the top of nearly every QRMI entry point) panics again, for the rest of the process's lifetime. So this isn't a rare, low-impact case; if it happens, QRMI becomes effectively unusable for the rest of that process.

Also fixed in [035213e](https://github.com/qiskit-community/qrmi/pull/188/commits/035213eaee935288aa6f697e06a6abce517f1793): I changed `common.rs`'s `initialize()` from `.init()` (panics on failure) to `.try_init() `(returns Result::Err on failure instead). This means losing the race with another library's logger setup now just silently disables QRMI's own log output -- `import qrmi` and every subsequent call keep working normally.

Net effect: moving initialization to import time does make this race trigger earlier, but switching to `.try_init()` makes losing that race harmless, so there should be no user-visible downside.

ohtanim: @awennersteen Do you agree with proceeding ?

awennersteen: Yes, i think so

PR #193, opened 2026-08-17
bugfix: `connection_verbose(true)` removed from release builds

### LINKED ISSUE #189: [Bug]: connection_verbose(true) leaks credentials (API keys, bearer tokens) into logs at RUST_LOG=trace

1. Build QRMI normally.
2. Set RUST_LOG=trace and run any script that makes a real request against a backend, e.g. examples/qiskit_primitives/ibm/sampler.py.
3. Observe the log output.

### What happens?

Several of the vendored HTTP clients under `dependencies/` unconditionally call `.connection_verbose(true)` on their `reqwest::Client::builder()`:

- `dependencies/qiskit_runtime_client/src/apis/configuration.rs`
- `dependencies/quantum_system_client/src/client.rs` (two builders)
- `dependencies/quantum_system_client/src/middleware/auth.rs`
- `dependencies/pasqal_cloud_client/src/client.rs`
- `dependencies/pasqal_local_client/src/client.rs`

`connection_verbose(true)` makes `reqwest` log the **raw bytes** written to and read from each connection at `TRACE` level, under the `reqwest::connect::verbose` target. This includes the literal HTTP request headers as sent on the wire -- in particular `Authorization` and any other credential-bearing header -- regardless of whether that header's `HeaderValue` was separately marked with `set_sensitive(true)` (see `quantum_system_client/src/client.rs`, which does call `set_sensitive(true)` on its auth header but is not protected by it here: `set_sensitive` only affects `HeaderValue`'s `Debug` output, not this wire-level byte dump, which already happened after headers were serialized).

At `RUST_LOG=trace`, this means API keys / bearer tokens end up verbatim in whatever destination QRMI's logs go to -- stderr, a file, or #188 potentially a Python `logging` handler or a user-supplied log callback.

Wire-level verbose logging should not be enabled in the builds we actually ship. Concretely:

- In debug builds (`cargo build` / `maturin develop`, which is debug by default), `connection_verbose(true)` can stay as-is; this is a developer running their own local build with their own credentials, and turning off the noisy default is the individual developer's call.
- In release builds (`maturin build --release`, `cargo build --release` --  what real releases are built from), `connection_verbose` should be `false` (reqwest's own default), so raw header/body dumps are never possible in a shipped build regardless of `RUST_LOG`.

This can be done at compile time with `#[cfg(debug_assertions)]` around each `.connection_verbose(true)` call, since `debug_assertions` already tracks exactly this debug/release distinction and doesn't require a new Cargo
feature or config surface.

```shell
2026-08-13 08:24:57,686 [DEBUG] reqwest::connect::verbose: write: b"GET /api/v1/backends/ibm_kingston/status HTTP/1.1\r\nauthorization: Bearer <REDACTED>\r\n...


(Redacted here; the actual output contains the full, live token.)
```

### Additional context

- Discovered while testing a separate change that bridges QRMI's Rust `log` output into Python's `logging` module (making previously-silent `RUST_LOG=trace` output actually visible when QRMI is used as a Python
  library) -- but the underlying leak is independent of that change and already exists on the C/CLI path today; the Python bridge only made it easier to notice.
- Severity: this exposes the caller's *own* credentials to their *own* log output, and only when they've explicitly opted into `RUST_LOG=trace`; it's not a remotely exploitable vulnerability. Still worth fixing, since `trace` is a reasonable thing to reach for while debugging an unrelated problem, and pasting logs (e.g. into a bug report, a chat, a CI log viewer) is an easy way to leak a token without realizing it.
- `set_sensitive(true)` is already used on at least one auth header (`quantum_system_client/src/client.rs`) with the evident intent of protecting it from appearing in logs; worth being aware that it does not cover this code path when deciding on a fix, so the fix should address `connection_verbose` directly rather than expanding `set_sensitive` usage.

badtst:
## Description of Change

Removed `connection_verbose(true)` from `reqwest` client builder in dependencies. This option could leak  authorisation bearing headers when setting `RUST_LOGS=trace`. 

Additional leaky `debug!` statements removed.

## Checklist ✅

## Ticket
- [x] Fixes #189 
- [ ] Is Part of #

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -220,6 +220,14 @@ is the property of your employer, then you will more than likely need to sign a
 [corporate CLA](https://qisk.it/corporate-cla) too and
 email it to us at <qiskit@us.ibm.com>.
 
+## Debug builds
+
+All `make build*`/`lint*`/`test*` targets build in release mode by default.
+Pass `DEBUG=1` (e.g. `DEBUG=1 make build`) to build in debug mode instead.
+See `make help` for details.
+
+For the python build process, run `maturin build` without the `--release` flag instead.
+
 ## Testing
 
 Once you've made a code change, it is important to verify that your change
@@ -265,10 +273,6 @@ documentation's [guide on writing tests](https://doc.rust-lang.org/book/ch11-01-
 For executing the tests with the makefile, a `make test` target is available.
 The execution of the tests (both via the make target and during manual
 invocation) takes into account the `LOG_LEVEL` environment variable.
-
-All `make build*`/`lint*`/`test*` targets build in release mode by default.
-Pass `DEBUG=1` (e.g. `DEBUG=1 make build`) to build in debug mode instead.
-See `make help` for details.
 
 
 ### Unsafe code and Miri
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -266,6 +266,10 @@ For executing the tests with the makefile, a `make test` target is available.
 The execution of the tests (both via the make target and during manual
 invocation) takes into account the `LOG_LEVEL` environment variable.
 
+All `make build*`/`lint*`/`test*` targets build in release mode by default.
+Pass `DEBUG=1` (e.g. `DEBUG=1 make build`) to build in debug mode instead.
+See `make help` for details.
+
 
 ### Unsafe code and Miri
 
--- a/dependencies/pasqal_cloud_client/src/client.rs
+++ b/dependencies/pasqal_cloud_client/src/client.rs
@@ -187,7 +187,9 @@ impl Client {
 impl Client {
     fn build_http_client() -> Result<reqwest_middleware::ClientWithMiddleware> {
         let mut reqwest_client_builder = reqwest::Client::builder();
-        reqwest_client_builder = reqwest_client_builder.connection_verbose(true);
+        if cfg!(debug_assertions) {
+            reqwest_client_builder = reqwest_client_builder.connection_verbose(true);
+        }
         let reqwest_builder = ReqwestClientBuilder::new(reqwest_client_builder.build()?);
         Ok(reqwest_builder.build())
     }
--- a/dependencies/pasqal_cloud_client/src/client.rs
+++ b/dependencies/pasqal_cloud_client/src/client.rs
@@ -238,7 +238,6 @@ impl Client {
     async fn handle_request<T: DeserializeOwned>(&self, resp: reqwest::Response) -> Result<T> {
         if resp.status().is_success() {
             let json_text = resp.text().await?;
-            debug!("{}", json_text);
             let val = serde_json::from_str(&json_text)?;
             Ok(val)
         } else {
--- a/dependencies/qiskit_runtime_client/src/apis/configuration.rs
+++ b/dependencies/qiskit_runtime_client/src/apis/configuration.rs
@@ -36,12 +36,13 @@ impl Configuration {
 
 impl Default for Configuration {
     fn default() -> Self {
-        let client = reqwest::Client::builder()
+        let mut client_builder = reqwest::Client::builder();
+        if cfg!(debug_assertions) {
             // Enabling this option will emit log messages at the TRACE level for
             // read and write operations on connections.
-            .connection_verbose(true)
-            .build()
-            .unwrap();
+            client_builder = client_builder.connection_verbose(true)
+        }
+        let client = client_builder.build().unwrap();
 
         Configuration {
             base_path: "/api".to_owned(),
--- a/dependencies/quantum_system_client/src/client.rs
+++ b/dependencies/quantum_system_client/src/client.rs
@@ -476,8 +476,10 @@ impl ClientBuilder {
             }
         }
 
-        reqwest_client_builder = reqwest_client_builder.connection_verbose(true);
-        reqwest_plain_client_builder = reqwest_plain_client_builder.connection_verbose(true);
+        if cfg!(debug_assertions) {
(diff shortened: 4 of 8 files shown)


dependencies/pasqal_cloud_client/src/client.rs
```diff
@@ -187,7 +187,9 @@ impl Client {
 impl Client {
     fn build_http_client() -> Result<reqwest_middleware::ClientWithMiddleware> {
         let mut reqwest_client_builder = reqwest::Client::builder();
-        reqwest_client_builder = reqwest_client_builder.connection_verbose(true);
+        if cfg!(debug_assertions) {
```
badtst, line 190: Did not use `#[cfg(debug_assertions)]` as attributes on expressions are experimental and would not compile. Only applies to whole items/statements

CONTRIBUTING.md
```diff
@@ -266,6 +266,10 @@ For executing the tests with the makefile, a `make test` target is available.
 The execution of the tests (both via the make target and during manual
 invocation) takes into account the `LOG_LEVEL` environment variable.
 
+All `make build*`/`lint*`/`test*` targets build in release mode by default.
+Pass `DEBUG=1` (e.g. `DEBUG=1 make build`) to build in debug mode instead.
+See `make help` for details.
+
```
ohtanim: In the Python build process, debug_assertions can be enabled by running maturin build without the --release flag. Should we mention this somewhere in the documentation?

dependencies/qiskit_runtime_client/src/apis/configuration.rs
```diff
@@ -36,12 +36,13 @@ impl Configuration {
...
+        let mut client_builder = reqwest::Client::builder();
+        if cfg!(debug_assertions) {
             // Enabling this option will emit log messages at the TRACE level for
             // read and write operations on connections.
-            .connection_verbose(true)
-            .build()
-            .unwrap();
+            client_builder = client_builder.connection_verbose(true)
+        }
+        let client = client_builder.build().unwrap();
```
ohtanim, line 45: I have tested those lines of changes. They worked as expected.

dependencies/quantum_system_client/src/client.rs
```diff
@@ -476,8 +475,10 @@ impl ClientBuilder {
             }
         }
 
-        reqwest_client_builder = reqwest_client_builder.connection_verbose(true);
-        reqwest_plain_client_builder = reqwest_plain_client_builder.connection_verbose(true);
+        if cfg!(debug_assertions) {
+            reqwest_client_builder = reqwest_client_builder.connection_verbose(true);
+            reqwest_plain_client_builder = reqwest_plain_client_builder.connection_verbose(true);
+        }
```
ohtanim, line 481: I have tested those lines of code. I noticed that I should remove another line to completely disable verbose log. Fixed in [b4c3cda](https://github.com/qiskit-community/qrmi/pull/193/commits/b4c3cda851e8135a7729129cd67143008918081f).

dependencies/quantum_system_client/src/middleware/auth.rs
```diff
@@ -51,7 +51,9 @@ impl TokenManager {
         retry_policy: Option<reqwest_retry::policies::ExponentialBackoff>,
     ) -> Result<Self> {
         let mut reqwest_client_builder = reqwest::Client::builder();
-        reqwest_client_builder = reqwest_client_builder.connection_verbose(true);
+        if cfg!(debug_assertions) {
+            reqwest_client_builder = reqwest_client_builder.connection_verbose(true);
+        }
```
ohtanim, line 56: I have tested those lines of code. I noticed that I should remove another line to completely disable verbose log. Fixed in [b4c3cda](https://github.com/qiskit-community/qrmi/pull/193/commits/b4c3cda851e8135a7729129cd67143008918081f).

dependencies/quantum_system_client/src/middleware/auth.rs
```diff
@@ -206,7 +208,6 @@ impl Middleware for AuthMiddleware {
         let token = token_manager.get_token().await?;
         // add authentication header to the request
         let mut cloned_req = request.try_clone().unwrap();
-        debug!("current token {}", token);
```
ohtanim, line 209: I remove this line to completely disable verbose log.

dependencies/quantum_system_client/src/middleware/auth.rs
```diff
@@ -225,7 +226,6 @@ impl Middleware for AuthMiddleware {
             debug!("renew access token");
             token_manager.get_access_token().await?;
             let token = token_manager.get_token().await?;
-            debug!("new token {}", token);
```
ohtanim, line 228: I remove this line to completely disable verbose log.

dependencies/quantum_system_client/src/client.rs
```diff
@@ -134,7 +134,6 @@ impl Client {
                 let status = resp.status();
                 if status.is_success() {
                     let json_text = resp.text().await?;
-                    debug!("{}", json_text);
```
ohtanim, line 137: I remove this line to completely disable verbose log.

ohtanim:
@badtst I have completed testing against IBM modules. You can proceed with this PR if Pasqal parts are verified.

CONTRIBUTING.md
badtst: To me debug build information should be constrained to `CONTRIBUTING.md` as they are only of interest to developers of the QRMI. However `CONTRIBUTING.md` does not have a section about python builds yet. So I would not document any of this in the python installation information of `INSTALL.md`. Maybe move this DEBUG build info to a dedicated section of `CONTRIBUTING.md` which would also include infor on maturin build ?

ohtanim (approved):
LGTM

PR #197, opened 2026-08-18
feat: add get_version() to the API

### LINKED ISSUE #196: Add `get_version()` to the API

The caller may have unexpected behaviour if the caller and libqrmi are no longer compatible after a system upgrade/downgrade.

Add an API like `get_version()` to allow callers to detect API compatibility issues at runtime.

API versions below version 1.0.0 are considered unstable - i.e. version zero. So checking API compatibility will make sense only after the QRMI API is bumped to version 1 - more details in #168 . 

_No response_

_No response_

Osheun:
Exposes the library version at runtime so callers can detect an incompatible `libqrmi` after a system upgrade or downgrade, as described in the issue.

The version comes from `env!("CARGO_PKG_VERSION")`, so it is baked in at compile time from `Cargo.toml` and cannot drift from the published package version.

### Three surfaces

| Surface | Signature |
|---|---|
| Rust | `qrmi::get_version() -> &'static str` |
| C | `const char *qrmi_get_version(void)` |
| Python | `qrmi.get_version() -> str` |

The C function returns a pointer to a static NUL-terminated string built with `concat!(env!("CARGO_PKG_VERSION"), "\0")`. It never allocates and never returns NULL, so there is nothing for the caller to free — the doc comment says so explicitly, since every other string-returning C function here does require `qrmi_string_free()` and I did not want that asymmetry to be a surprise.

The Python function is registered on the `_core` module and reaches users through the existing `from ._core import *` in `python/qrmi/__init__.py`, so no extra plumbing was needed.

### Verification

- `cargo build --lib` — clean
- `cargo check --lib --features pyo3` — clean
- `cargo clippy --lib` — no new warnings
- `cargo test --doc get_version` — the doctest on the Rust function passes
- cbindgen emits `const char *qrmi_get_version(void);` into the generated `qrmi.h` as expected

### One thing to confirm

I tagged the C function `@version 0.23.0`, guessing the next minor after the current `0.22.0`. Happy to change it if the release it lands in ends up being different.

As #168 notes, compatibility checks only really become meaningful once the API reaches 1.0 — this just puts the primitive in place so callers can start relying on it.

Closes #196

### DIFF
--- a/src/cext.rs
+++ b/src/cext.rs
@@ -261,6 +261,29 @@ pub unsafe extern "C" fn qrmi_log_callback_set(callback: QrmiLogCallback) -> Ret
 }
 
 /// @ingroup Qrmi
+/// Returns the version of this QRMI library as a semantic version string.
+///
+/// Callers can compare this against the version they were built against to detect
+/// incompatibilities at runtime, for example after a system upgrade replaced the
+/// shared library underneath them.
+///
+/// The returned pointer refers to a string with static lifetime owned by the
+/// library. It must NOT be passed to `qrmi_string_free()`.
+///
+/// # Example
+///
+/// @code
+///   printf("QRMI version = %s\n", qrmi_get_version());
+/// @endcode
+///
+/// @return pointer to a NUL-terminated version string such as "0.22.0". Never NULL.
+/// @version 0.23.0
+#[no_mangle]
+pub extern "C" fn qrmi_get_version() -> *const c_char {
+    concat!(env!("CARGO_PKG_VERSION"), "\0").as_ptr() as *const c_char
+}
+
+/// @ingroup Qrmi
 /// Free a string allocated by C API
 ///
 /// # Safety
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,6 +29,25 @@ use crate::models::{Payload, ResourceType, Target, TaskResult, TaskStatus};
 use anyhow::Result;
 use async_trait::async_trait;
 
+/// Version of this QRMI library, as a semantic version string such as `"0.22.0"`.
+///
+/// Callers can compare this against the version they were built against to detect
+/// incompatibilities at runtime, for example after a system upgrade replaced the
+/// shared library underneath them.
+///
+/// Note that versions below `1.0.0` are considered unstable, so a mismatch in the
+/// minor component is enough to signal a potentially breaking change.
+///
+/// # Example
+///
+/// ```
+/// let version = qrmi::get_version();
+/// println!("{version}"); // prints e.g. "0.22.0"
+/// ```
+pub fn get_version() -> &'static str {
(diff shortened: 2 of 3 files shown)


awennersteen:
Thank you for contributing @Osheun! I've been off for some weeks so I'm not up to date yet on what has happened, but I know this was a very actively moving piece.

@ohtanim Can you comment here?

ohtanim:
> @ohtanim Can you comment here?

Well, I don't see the need for being able to programmatically retrieve the crate version via `get_version()`. I can't think of a real use case that requires this.

The #225 I implemented is aligned with an actual use case and delivers real value in actual maintenance operations.

I tried to think about what someone would actually want to do with a value obtained programmatically via `get_version()`. It's probably one of the following two things:

1. Print it as a message in logs or on the console
2. Have the caller learn the version and change its own behavior depending on that version 

For the first case, there's no need to be able to know it programmatically — #225 already covers this.

For the second case, I'd expect code like:

```
if version == "0.24.0" or version == "0.24.1" or version == "0.24.2" .... :
      // do something
elif version == "0.25.0" :
      // do something
```

But this doesn't allow specifying something like `version >= 0.24.0`, and it forces whoever reads the code to understand what was implemented in each individual version. From my perspective, this is a very awkward interface to work with.

Rather than providing something like `get_version()`, I strongly recommend a `capabilities()` interface. As the name suggests, this interface returns what the system can actually do. For example, it could return JSON like the following:

1) QRMI-level

```
{
     "supported_providers": [
         "ibm-quantum-system",
         "ibm-quantum-compute",
         "pasqal-cloud",
         "pasqal-local",
         "alice-bob-felis",
         "iqm-server"
     ],
}
```

If it's absolutely necessary, the crate version could be included here too, but I doubt it would actually be used for anything.

2)QuantumResource level

```
{
     "supported_api_versions": [
         "1.0",
         "1.1"
     ],
     "is_resource_acquisition_supported": true,
     "is_dynamic_resource_discovery_supported": false,
     "num_qubits": 127,
     "quantum-type": "superconducting",
     "max_shots": 50000,
     "supported_payload_type": "QiskitPrimitive",
     "vendor-specific": {
         "supported_gates": [
             "cx",
             "rz",
             "rzz"
         ],
         "supported_instructions": [
             "xxx",
             "yyy",
             "zzz"
         ],
         "supported_program_ids": [
             "sampler",
             "estimator",
             "executor",
         ]
         ...
      }
}

```
I believe this would be far more useful for programmatic control than simply returning a version number.

@awennersteen What do you think?

Osheun:
Thanks for the detailed explanation @ohtanim, and sorry it took me so long to get back to this.

You're right. #225 already covers the logging/console case, and branching on version strings would be an awkward thing to encourage. A `capabilities()` interface answers the actual question much better than a version number does.

I'm closing this one. Thanks @awennersteen for looping in the right person.

PR #200, opened 2026-08-20
Introduce typed error handling (QrmiError) to replace anyhow across public APIs

ohtanim:
## Description of Change

Replaces `anyhow::Error` with a typed `QrmiError` (built with `thiserror`) across the public `QuantumResource` and `ResourceProvider` trait APIs, and threads that type through to the C and Python bindings so callers can match on specific
failure kinds instead of parsing error strings.

### Motivation

Previously, every fallible operation returned `anyhow::Result<T>`. This made it easy to add context to errors (`.context(...)`), but callers — in Rust, C, and Python — had no way to distinguish *why* a call failed short of string-matching the error message. This PR introduces a proper error type while keeping the migration low-risk for existing callers.

#### `QrmiError` (`src/error.rs`)
- New `thiserror`-based enum covering conditions generic to every backend: `EnvVarNotSet`, `ParseError`, `UnsupportedResourceType`, `UnsupportedPayload`, `TaskNotReady`, `MissingConfigKey`, `InvalidFilter`, `InvalidJson`,
  `InvalidUtf8`, and a catch-all `Other(anyhow::Error)`.
- `QrmiError::kind() -> QrmiErrorKind` gives a stable, `#[repr(C)]` classification tag independent of the message text, used by the C and Python bindings.

#### Vendor-specific errors
- Conditions that only make sense for one vendor (e.g. an unknown IBM program ID, an invalid Pasqal device type) live in their own types — `crate::ibm::error::IbmError`, `crate::pasqal::error::PasqalError` — rather than being folded into the framework-level `QrmiError`. `QrmiError::Ibm(...)` / `QrmiError::Pasqal(...)` wrap them with `#[from]`, so `?` still works end to end. Adding a new failure mode to an existing vendor never touches `error.rs`; only adding a new vendor does.

#### `QuantumResource` / `ResourceProvider` trait signatures
- Changed from `anyhow::Result<T>` to `crate::Result<T>` (`= std::result::Result<T, QrmiError>`) across all backends (IBM Qiskit Runtime Service, IBM Quantum System, IQM, Pasqal Cloud/Local, Alice & Bob Felis). 
- Every `bail!(...)` / `anyhow!(...)` call site in the vendor implementations was replaced with either a specific `QrmiError` variant or `.context("...")` where the underlying vendor client already returns its own error type.
- No blanket `impl<E: std::error::Error> From<E> for QrmiError` — this would conflict with the vendor-specific `#[from]` impls under Rust's coherence rules. Generic errors (JSON, UTF-8) get a dedicated variant; vendor API
  client errors go through `.context(...)`.

#### C bindings (`cext.rs`)
- `ReturnCode` gained variants matching `QrmiErrorKind` (e.g. `EnvVarNotSetError`, `TaskNotReadyError`).
- New `qrmi_get_last_error_kind()` returns a machine-readable classification alongside the existing `qrmi_get_last_error()` message string.

#### Python bindings (`pyext.rs`)
- New exception hierarchy rooted at `qrmi.QrmiError_` (subclassing `RuntimeError` for backward compatibility), with specific subclasses (`EnvVarNotSetError`, `ConfigError`, `UnsupportedResourceTypeError`,    `UnsupportedPayloadError`, `TaskNotReadyError`, `InvalidFilterError`) so Python callers can `except qrmi.TaskNotReadyError` instead of matching on message text.

### Backward compatibility

- `QrmiError` implements `std::error::Error`, so it still converts to `anyhow::Error` automatically via `?` — code calling into `qrmi` that itself returns `anyhow::Result` needs no changes.
- C callers using only `qrmi_get_last_error()` continue to work unchanged; `qrmi_get_last_error_kind()` is additive.
- Python callers catching the previous generic `RuntimeError` continue to work unchanged, since every new exception subclasses it.

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #


src/pyext.rs
```diff
@@ -25,6 +26,74 @@ use pyo3::prelude::*;
...
 use tokio::runtime::Runtime;
 
+pyo3::create_exception!(
+    qrmi,
+    QrmiError_,
+    pyo3::exceptions::PyRuntimeError,
+    "Base class for all QRMI-specific errors. Catching this catches any \
+     error QRMI itself raises (as opposed to errors surfaced verbatim from \
+     an underlying vendor library)."
+);
```
badtst: Exceptions are created here but not added to the module, so I can't import them directly like `from qrmi import ConfigError`. Maybe add to the `_core` pymodule

src/pyext.rs
```diff
@@ -25,6 +26,74 @@ use pyo3::prelude::*;
...
+        QrmiErrorKind::UnknownProgramId | QrmiErrorKind::UnsupportedPayload => {
+            UnsupportedPayloadError::new_err(msg)
+        }
+        QrmiErrorKind::TaskNotReady => TaskNotReadyError::new_err(msg),
+        QrmiErrorKind::InvalidFilter | QrmiErrorKind::InvalidValue => {
+            InvalidFilterError::new_err(msg)
+        }
+        QrmiErrorKind::Other => QrmiError_::new_err(msg),
+    }
+}
```
badtst, line 95: nit: Maybe explicit a little bit more the reasons for those exceptions folding ? Or add it to some documentation ?

badtst (commented):
Lgtm ! Minor issue with python bindings

src/pyext.rs
ohtanim: > Maybe explicit a little bit more the reasons for those exceptions folding ?

At the moment, I have listed the exceptions that are explicitly thrown by each vendor implementation. We expect additional exceptions, such as AuthError, to be introduced over time. That is why QrmiError is defined in a way that allows it to accommodate future extensions. Let's continue to enhance the framework-level error definitions.

src/pyext.rs
ohtanim: > Exceptions are created here but not added to the module, so I can't import them directly like `from qrmi import ConfigError`. Maybe add to the `_core` pymodule

@badtst Oh. Thank you. Good catch. I forgot to add the m.add() calls for those exception classes.
I committed src/pyext.rs file again. Exception structure will be like below. QrmiErrorKind::Other (=Vendor errors) in Rust will be mapped to QrmiError. Thanks, again.

```
BaseException
 └── Exception
      └── RuntimeError
           └── QrmiError (= QrmiError_)
                ├── EnvVarNotSetError
                ├── ConfigError
                ├── UnsupportedResourceTypeError
                ├── UnsupportedPayloadError
                ├── TaskNotReadyError
                └── InvalidFilterError
```

badtst (approved):
lgtm

PR #208, opened 2026-08-21
Reimplement QRMIService in Rust with Python and C bindings

### LINKED ISSUE #165: QRMIService in Rust

The QRMIService feature in Python is a useful UI tool that sets up all QuantumResources defined in env vars automatically but doesn't exist in our other supported languages. This tool makes using QRMI easier, simpler and removes some of the grunt work and env var interpretation from the user.

Move the QRMIService feature to Rust and then bind it through to supported languages. Implemented by:

- [x] struct defined as pub struct QRMIService {resources: HashMap<String, QuantumResource>}

- [x] Initiated as QRMIService:from_environment() -> Result<Self>

- [x] Access resources via .resources() -> Vec<&QuantumResource>

- [ ] Bind through to C, Python, Go*

- [x] Write tests

- [ ] Set up API docs

Any other suggested features or methods to add to this would be appreciated.

_No response_

_No response_

ohtanim:
## Description of Change

To reviewer,

I'm not sure what the current status of the implementation for #165 is. Since I was hoping to have this functionality available sooner, I went ahead and put together an implementation myself, including the Python/C bindings.This enhancement has actually been discussed since before ISC26, so I thought it would be worthwhile to provide a working implementation. That said, if someone else is working on or plans to contribute an implementation that provides equivalent or better functionality, I would be more than happy to withdraw this contribution. As this is only a relatively small enhancement that took me roughly two hours to implement, my primary goal is simply to see the functionality made available in the best form possible.

### Summary

Re-implements `QRMIService` in Rust (`qrmi::QRMIService`) and exposes it through the Python and C bindings, following the same pattern already used for `QuantumResource` and `ResourceProvider`. This does **not** change what `QRMIService` does or how it's used from Python -- the goal is to move the implementation, not the interface.

Before this PR, `QRMIService` only existed as pure Python (`python/qrmi/primitives/service.py`), calling into the Rust-backed `QuantumResource`/`ResourceType` bindings underneath. After this PR, the discovery/filtering logic itself lives in Rust, and Python is a thin pyo3 wrapper around it -- the same relationship `QuantumResource` and `ResourceProvider` already have between their Rust and Python layers.

### Motivation

- Consistency: `QuantumResource` and `ResourceProvider` are both implemented in Rust with Python/C bindings on top; `QRMIService` was the one exception, implemented directly in Python. This PR closes that gap.
- Reuse: `QRMIService`'s discovery logic (reading `QRMI_JOB_QPU_RESOURCES`/`QRMI_JOB_QPU_TYPES`, filtering by
  accessibility) is now available to Rust and C callers, not just Python.

#### Rust core: `qrmi::QRMIService` (new)

New module `src/service.rs`, exported at the crate root (`pub use service::QRMIService`). Discovers the QPU resources assigned to the current job 
-- read from `QRMI_JOB_QPU_RESOURCES` / `QRMI_JOB_QPU_TYPES` (or the legacy `SLURM_JOB_QPU_RESOURCES` / `SLURM_JOB_QPU_TYPES`) 
-- and keeps only the ones that are currently accessible.

```rust
impl QRMIService {
    pub async fn new() -> Result<Self>;

    /// Borrowing accessors: repeated calls return handles into the same
    /// underlying resources (not fresh instances).
    pub fn resources(&mut self) -> Vec<&mut (dyn QuantumResource + Send + Sync + 'static)>;
    pub fn resource(&mut self, resource_id: &str) -> Option<&mut (dyn QuantumResource + Send + Sync + 'static)>;
}
```

Also refactors the `ResourceType` <-> string mapping (`ibm-quantum-system`, `ibm-quantum-compute-service`, `qiskit-runtime-service`, `pasqal-cloud`,
`pasqal-local`, `alice-bob-felis`, `iqm-server`) into a single `ResourceType::from_qpu_type_str()` in `src/models/config.rs`, used by both `QRMIService::new()` and the existing config-file `Deserialize` impl (previously duplicated inline in a couple of places).

#### Python bindings: `qrmi.QRMIService`

`PyQRMIService` in `src/pyext.rs` is a thin wrapper around `qrmi::QRMIService`: it drives `QRMIService::new()` to completion on a private tokio runtime, then moves each discovered resource into its own, independently owned `PyQuantumResource` (the same pattern `PyResourceProvider::resources()` already uses for `ResourceProvider::resources()`).

**The Python-visible interface is unchanged.** `python/qrmi/primitives/service.py` no longer contains the implementation -- it's now a two-line re-export of `qrmi.QRMIService` -- but every existing import path keeps working exactly as before, with no deprecation:

```python
from qrmi import QRMIService                    # new, direct path
from qrmi.primitives import QRMIService          # unchanged
from qrmi.primitives.service import QRMIService  # unchanged (existing user code)
```

All three resolve to the same class. `QRMIService()`, `.resources()`, and `.resource(id)` keep their existing signatures and behavior, including returning the *same* `QuantumResource` instance on repeated `.resource(id)` calls (e.g. after `.acquire()`).

One intentional behavior difference: the length-mismatch check between `QRMI_JOB_QPU_RESOURCES` and `QRMI_JOB_QPU_TYPES` now raises `RuntimeError` instead of `ValueError`, to match how every other error in the Rust
bindings is surfaced to Python. Flagging this explicitly in case it matters to anyone catching `ValueError` specifically.

#### C bindings (new)

```c
QrmiReturnCode qrmi_service_resources(QrmiQuantumResources *resources_out);
QrmiReturnCode qrmi_service_resources_free(QrmiQuantumResources *resources);
```

No new `#[repr(C)]` types -- reuses the existing `QrmiQuantumResources` / `QrmiQuantumResource` from the `ResourceProvider` C API.

Also fixes a pre-existing gap in `cbindgen.toml`: `@ingroup QrmiResourceProvider` was used throughout `cext.rs` but the corresponding `@defgroup QrmiResourceProvider` was never defined. Added it alongside the new `@defgroup QrmiService`.

#### Examples

- `examples/qrmi/rust/service/` -- uses `qrmi::QRMIService` directly.
- `examples/qrmi/python/service/` -- uses `qrmi.QRMIService` directly.
- `examples/qrmi/c/service/` -- uses `qrmi_service_resources()`.

Each lists accessible resources, and if a resource id is passed on the command line, additionally acquires it, prints its metadata/target, then releases it.

### Backward compatibility

- **Python:** No changes required in downstream code. All existing import paths and the class's public interface are unchanged.
- **Rust/C:** Purely additive (new module, new `#[no_mangle]` functions).

### Testing

- [x] `cargo build` / `cargo build --features pyo3` / `cargo clippy -- -D warnings` (all feature combinations)
- [x] `make build-stubgen` -- confirm `QRMIService` shows up correctly in the generated `.pyi`
- [x] `pytest python/tests`
- [x] `cargo build --release --example qrmi-example-service`
- [x] C example builds and links against the freshly-built `libqrmi` (`examples/qrmi/c/service`)
- [x] `python examples/qrmi/python/service/example.py` against a real job environment
- [x] `python examples/qiskit_primitives/ibm/sampler.py` against a real job environment
- [x] `cbindgen`-generated header includes `qrmi_service_resources` / `qrmi_service_resources_free` under the `QrmiService` Doxygen group

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #


awennersteen (commented):
Year

examples/qrmi/c/service/src/common.c
```diff
@@ -0,0 +1,66 @@
+/*
+ * This code is part of Qiskit.
+ *
+ * (C) Copyright IBM 2025.
```
awennersteen: Year

examples/qrmi/c/service/src/common.c
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Any modifications or derivative works of this code must retain this
+ * copyright notice, and modified files need to carry a notice indicating
+ * that they have been altered from the originals.
+ */
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+void load_dotenv() {
```
awennersteen, line 18: Can we not do this in rust? To keep a single implementation?

ohtanim: We can.

examples/qrmi/c/service/src/common.c
ohtanim: Fixed.

examples/qrmi/c/service/src/common.c
ohtanim: @awennersteen I've reconsidered this. If needed, I think it would be better to address this change in a separate PR. For this PR, I'm focusing on implementing the same functionality that we currently have. Do you agree?

awennersteen: Yep, makes sense

ohtanim:
@awennersteen @MatthieuMoreau0 @badtst I've pulled in the latest changes from main, adjusted the code where necessary, and verified everything by rerunning the tests. Ready to restart to review again.

ohtanim:
I will wait until the end of today for the review and approval. However, if they are not provided, I will take ownership of the risk and proceed with the merge, given that this enahcement does not affect existing functionality, is a simple feature that can be implemented quickly(=2 hours work), and has already been reviewed by AI.

In parallel, I am hoping that other implementations will emerge as well.It appears that the STFC team is working on a [similar enhancement](https://github.com/qiskit-community/qrmi/tree/qrmi_service), but their effort seems to be limited to the Rust API, with no work yet started on the C or Python bindings. A Rust implementation alone is not particularly useful, as it cannot realistically be used without those bindings.

MatthieuMoreau0:
Hi @ohtanim, apologies for the late feedback. I have already reviewed the code and it looks good to me; I am currently testing it against our end to end test suite but I have been fighting some issues with my local setup. I think I almost sorted it out and will approve as soon as I get a successful run on my side

MatthieuMoreau0 (approved):
LGTM

PR #212, opened 2026-08-22
[IBM] quantum_system_client: replace anyhow with QuantumSystemError, fix retry-related bugs found along the way

ohtanim:
## Description of Change

@yoonho Could you please review and approve ? This PR updates the IBM Quantum System API and QRMI to support the recently introduced typed error enhancements #200 . This work is part of our response to feedback from openQSE. Although many files are affected, the changes are straightforward and mainly consist of replacing generic anyhow::Result usage with appropriate QuantumSystemError mappings.

@awennersteen Added `ResourceNotFound`, `TaskNotFound` and `AuthenticationFailed` to QRMI errors. fyi.

### Summary

Replaces `anyhow::Result`/`anyhow::Error` with a typed `QuantumSystemError` (built with `thiserror`) across every public `Client`/`PrimitiveJob` method in this crate, and fixes several retry-related bugs uncovered while wiring up
error classification end-to-end.

### Motivation

Every fallible operation previously returned `anyhow::Result<T>`. This made it easy to attach context via `.context(...)`, but callers (in this crate's consumer, `qrmi`) had no way to distinguish *why* a call failed short of
string-matching the message — e.g. "the backend doesn't exist" and "the request timed out" were indistinguishable at the type level. This PR introduces a proper error type so callers can match on the actual failure mode instead.

#### `QuantumSystemError` (`src/error.rs`)
- New `thiserror`-based enum mapping well-known HTTP status codes onto specific variants: `BackendNotFound`/`JobNotFound` (404, disambiguated via a `ResourceKind` parameter passed at the call site — the status code alone doesn't say which), `AuthenticationFailed` (401), `AccessDenied` (403), `RequestTimeout` (408), `InvalidJobInput` (400/409/413/422), `ExecutionLanesFull` (429), `ServerError` (5xx), and a generic `Api {
  status, body }` fallback for anything else. 
- `Request(String)` covers requests that never got a response at all (connection failure, DNS, TLS, ...).
- `Reqwest`/`Json`/`Io` wrap `reqwest::Error`/`serde_json::Error`/ `std::io::Error` directly via `#[from]`.
- `Other { message, source }` is the catch-all for lower-level SDK errors (e.g. `aws-sdk-s3`) that don't warrant their own variant, still carrying the original error as `source` for anyone inspecting the chain.
- `from_response()` and `from_middleware_error()` centralize the response/error → variant mapping so each of the ~20 call sites doesn't repeat it.

#### Every API call site (`src/api/*.rs`, `src/client.rs`)
- Every `bail!`/`anyhow!` was replaced with a specific `QuantumSystemError` variant; every bare `?` on an external error type now goes through the new `#[from]` conversions or an explicit `.map_err(...)`.
- `middleware/auth.rs`'s `Middleware` trait impl still builds errors with `anyhow!` for the reqwest_middleware::Error::Middleware` variant, since that's dictated by an external crate's type and isn't something we control.

#### Propagation into `qrmi::QrmiError` (`src/error.rs`, `src/ibm/error.rs`)
- `QrmiError` gained three new variants: `ResourceNotFound`, `TaskNotFound`, and `AuthenticationFailed` (same name), plus matching `QrmiErrorKind` entries.
- A new `impl From<quantum_system_api::QuantumSystemError> for QrmiError` in `ibm/error.rs` maps `BackendNotFound → ResourceNotFound`, `JobNotFound → TaskNotFound`, `AuthenticationFailed → AuthenticationFailed` (same name), and everything else into `QrmiError::Other` (still carrying
  the original error as `source`).
- `ibm/quantum_system.rs` and `ibm/quantum_system_provider/mod.rs`'s `.context("...")?` calls into `quantum_system_api` were replaced with plain `?`, since `.context()` would otherwise wrap the error in
  `anyhow::Error` *before* the new `From` conversion gets a chance to run, collapsing everything back into `QrmiError::Other` regardless of the actual cause.

#### C and Python bindings (`src/cext.rs`, `src/pyext.rs`)
- `ReturnCode` gained `ResourceNotFoundError`, `TaskNotFoundError`, and `AuthenticationFailedError` (appended after the existing codes; no renumbering).
- `pyext.rs` gained matching `qrmi.ResourceNotFoundError`, `qrmi.TaskNotFoundError`, and qrmi.AuthenticationFailedError` exception classes, registered in both `to_py_err()`'s mapping and the module's
  `m.add(...)` calls.

#### Retry-related bugs found along the way
Wiring up accurate error classification surfaced three separate, pre-existing retry bugs:

1. **`reqwest_retry::RetryTransientMiddleware` erases error type information.** Even after zero retries, it unconditionally rewraps whatever error it receives into an opaque `RetryError`, discarding the `source()` chain. This made it impossible to reliably tell "IAM rejected the credentials" apart from "a plain DNS failure occurred" once wrapped. Replaced with `middleware/retry.rs`'s `TransparentRetryMiddleware`: same retry *decision* logic (reuses the same `RetryPolicy`/`RetryableStrategy` traits, so behavior is unchanged), but returns the final result unmodified instead of rewrapping it.

2. **Retry multiplication from middleware ordering.** The retry middleware was attached *before* (i.e. outside) `AuthMiddleware`. A transient failure on the actual request caused the *whole* `AuthMiddleware::handle()`
   call to be retried — including redundantly re-running its own internal, already-retried token acquisition every time — turning what should be a handful of attempts into a multiplicative explosion (outer retries × inner retries). Fixed by attaching `AuthMiddleware` first (outermost) and the retry middleware second (innermost, wrapping only the actual request send).
3. **`AuthMiddleware` retried-with-a-fresh-token on *any* error, not just 401.** `response.is_err() || status == 401` triggered a full manual retry (including re-running the entire inner retry chain) for connection timeouts and DNS failures that have nothing to do with token freshness. Narrowed to only retry when the actual API responded with 401.
4. **The `iqp_retry_policy` feature's 429-specific retry strategy also caught generic network failures.** `RetryStrategy429`'s `Err` arm delegated to `default_on_request_failure`, so a connection failure would
   also be retried under `retry_policy_429`'s *unlimited* retry count (appropriate for waiting out a rate limit, not for redialing an address that will never resolve). Narrowed to only handle 429/423; everything else now falls through to the normal, bounded retry layer.

### Backward compatibility

This is a breaking change for `quantum_system_client`'s public API: every method's `Result<T>` error type changes from `anyhow::Error` to `QuantumSystemError`. `QuantumSystemError` implements `std::error::Error`,
so it still converts to `anyhow::Error` automatically via `?` for any consumer that itself uses `anyhow::Result`.

The consuming `qrmi` crate has been updated in a companion change: its own `QrmiError` gained `ResourceNotFound`/`TaskNotFound`/`AuthenticationFailed` variants, and `ibm/quantum_system.rs`'s `.context(...)` calls into this crate were replaced with plain `?` (relying on a new `From<QuantumSystemError> for QrmiError` conversion) so the richer classification isn't lost at that boundary.

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2623,6 +2623,7 @@ dependencies = [
  "reqwest",
  "reqwest-middleware",
  "reqwest-retry",
+ "retry-policies",
  "rhexdump",
  "serde",
  "serde_json",
--- a/dependencies/quantum_system_client/Cargo.toml
+++ b/dependencies/quantum_system_client/Cargo.toml
@@ -26,6 +26,7 @@ async-trait = "0.1.83"
 http = "1.1.0"
 reqwest-middleware = { version = "0.4.2", features = ["json"] }
 reqwest-retry = "0.7.0"
+retry-policies = "0.4"
 tokio = { version = "1.40.0", features = ["full"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/dependencies/quantum_system_client/src/api/backend_config.rs
+++ b/dependencies/quantum_system_client/src/api/backend_config.rs
@@ -10,7 +10,7 @@
 // that they have been altered from the originals.
 
 use crate::Client;
-use anyhow::Result;
+use crate::Result;
 use serde::de::DeserializeOwned;
 
 impl Client {
@@ -19,7 +19,6 @@ impl Client {
     /// # Example
     ///
     /// ```no_run
-    /// use anyhow::Result;
     /// use quantum_system_api::{AuthMethod, ClientBuilder};
     ///
     /// #[tokio::main]
@@ -53,6 +52,7 @@ impl Client {
             "{}/v1/backends/{}/configuration",
             self.base_url, backend_name
         );
-        self.get::<T>(&url, &self.client).await
+        self.get::<T>(&url, &self.client, crate::error::ResourceKind::Backend)
+            .await
     }
 }
--- a/dependencies/quantum_system_client/src/api/backend_details.rs
+++ b/dependencies/quantum_system_client/src/api/backend_details.rs
@@ -10,7 +10,7 @@
 // that they have been altered from the originals.
 
 use crate::Client;
-use anyhow::Result;
+use crate::Result;
 use serde::de::DeserializeOwned;
 
 impl Client {
@@ -46,6 +46,7 @@ impl Client {
     /// - specified backend is not found.
     pub async fn get_backend<T: DeserializeOwned>(&self, backend_name: &str) -> Result<T> {
         let url = format!("{}/v1/backends/{}", self.base_url, backend_name);
-        self.get::<T>(&url, &self.client).await
+        self.get::<T>(&url, &self.client, crate::error::ResourceKind::Backend)
+            .await
     }
 }
--- a/dependencies/quantum_system_client/src/api/backend_lanes_config.rs
+++ b/dependencies/quantum_system_client/src/api/backend_lanes_config.rs
@@ -10,7 +10,7 @@
 // that they have been altered from the originals.
 
 use crate::Client;
-use anyhow::Result;
+use crate::Result;
 use serde::de::DeserializeOwned;
 
 impl Client {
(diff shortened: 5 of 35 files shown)


yoonho (approved):
Very nice! Thank you for addressing Doug's feedback so quickly

PR #213, opened 2026-08-23
[IQM] Classify IQM Server API errors, consolidate QrmiError's "invalid input" variants

ohtanim:
## Description of Change

@yoonho This PR addresses IQM QRMI changes to add more typed error for openQSE comments. Could you please review and approve. Thanks,

### Summary

Extends the typed-error work done in #212  (`quantum_system_client`) to the IQM backend, and along the way simplifies `QrmiError`'s taxonomy: what were three separate "something was invalid" variants (`InvalidFilter`,
`InvalidValue`, and a short-lived `InvalidRequest`) are consolidated into one `QrmiError::InvalidInput`. While testing this against a live server, also found and fixed two pre-existing bugs unrelated to error classification: a response-shape mismatch in the generated `get_qc_health_v1` client, and `task_stop` unconditionally trying to cancel jobs that had already finished.

### Motivation

`src/iqm/server.rs` wrapped every call into `iqm_server_api` with `anyhow::Context`, so every IQM failure surfaced as `QrmiError::Other` regardless of cause -- a 401 looked the same as a 404 or a network timeout. Unlike `quantum_system_client` (#212 ), `iqm_server_api` is OpenAPI-generated code we don't own or modify, so it needed a different approach.

Separately, while wiring that up, it became clear that `QrmiError` had accumulated more granularity than it needed: `InvalidFilter` (a malformed `filters` string), `InvalidValue` (JSON/UTF-8/vendor-specific bad values), and `InvalidRequest` (a vendor's API rejecting a request after receiving it) all meant the same thing to a caller -- "the input was bad" -- and in practice a caller ends up checking for all of them together anyway. They're now one variant.

#### IQM error classification (`src/iqm/error.rs`, new)
- `iqm_server_api` generates a *different* error enum per operation (`GetJobV1Error`, `CancelJobV1Error`, ...), so there's no single type to hang a shared classifier off of the way `quantum_system_client` could. `classify::<T>()` instead works off `Error::ResponseError`'s `status` field, which every operation's error carries regardless of the per-operation type `T`.
- Maps HTTP status code:
  -  401 → `AuthenticationFailed` (uniform across every endpoint -- always the `Unauthorized` model)
  -  400 → `InvalidInput` (uniform too -- `InvalidInput`, or for `job_submit` specifically one of  `InvalidJobPayload`/`UnsupportedJobType`/`InvalidInput`)
  -  404 → `ResourceNotFound`/`TaskNotFound` depending on a `ResourceKind` parameter passed at each call site (the status code alone doesn't say whether it was a job or a quantum computer that wasn't found).
- Deliberately does *not* classify 403: it means something different per endpoint in this API (e.g. `cancel_job_v1`'s 403 is `IllegalJobStatus` -- the job's state doesn't allow cancelling it, nothing to do with auth), so guessing would be wrong more often than not. Falls through to `Other`, still carrying the original error (with full `source()` chain) rather
  than losing it.
- `src/iqm/server.rs`: every `.context(...)?`/`.with_context(...)?` replaced with `.map_err(|e| classify(e, ResourceKind::...))?`; JSON-decode failures now go through the existing `QrmiError::InvalidJson` via a bare `?`
  instead of `anyhow::Context`, since that no longer needs a message wrapper.

#### `QrmiError` taxonomy cleanup (`src/error.rs`)
- `InvalidFilter`, `InvalidValue`, and `InvalidRequest` are gone; everything that used to construct one of them now constructs `QrmiError::InvalidInput(String)`, with a single `QrmiErrorKind::InvalidInput`.
- `quantum_system_client::QuantumSystemError::InvalidJobInput` (400/409/413/422, from #212 ) now maps to `QrmiError::InvalidInput` via `ibm/error.rs` -- previously it fell through to `Other` unnoticed.
- `IbmError::UnknownProgramId`/`InvalidSessionMode` and  `PasqalError::InvalidDeviceType`/`InvalidCudaqSequence` still exist as their own vendor-level variants (Rust-level matching stays precise), but now tag as `QrmiErrorKind::InvalidInput` like everything else.
- `InvalidJson`/`InvalidUtf8` are unchanged as distinct `QrmiError` variants (they need to stay separate for their `#[from]` conversions to work with `?`), but also now tag as `InvalidInput`.

#### C and Python bindings
- `cext.rs`: `ReturnCode::InvalidRequestError`/`InvalidValueError` (109/110, introduced earlier in this same line of work) collapse into a single `InvalidInputError` (109); 110 is retired rather than reused, since nothing has shipped against these codes yet.
- `pyext.rs`: `qrmi.InvalidFilterError`/`InvalidValueError`/`InvalidRequestError` collapse into `qrmi.InvalidInputError`, registered in both `to_py_err()` and the module's `m.add(...)` calls.

#### Two bugs found while testing against a live server

Wiring up real error classification meant actually exercising every code path against a live IQM Server instance, which surfaced two pre-existing bugs unrelated to error handling itself:

- **`get_qc_health_v1`'s response shape didn't match the live server(Seems recently changed..).**
  The generated `IqmServerQcHealthStatus` model expected a flat  `{"healthy": ..., "updated_at": ...}` body, but the server actually returns `{"operational": "...", "health": {"healthy": ..., "updated_at": ...}}` -- `healthy` one level deeper than the generated client expects. `is_accessible()` was failing outright with a `missing field 'healthy'`
  deserialization error as a result. Since `dependencies/iqm_client` is generated code, hand-patched `IqmServerQcHealthStatus` (adding a new nested `QcHealthDetail` struct for the `health` object) rather than
  editing `get_qc_health_v1` itself, which deserializes the whole response directly into that struct and so didn't need to change. The patch is flagged in a doc comment as something to drop once the crate is regenerated from a spec that reflects the server's current shape. `is_accessible()` updated accordingly (`health.healthy` →
  `health.health.healthy`).
- **`task_stop` unconditionally tried to cancel every job**, including ones that had already reached a terminal state, which the server correctly rejects with 403 (`IllegalJobStatus`) -- visible as log noise in the C example even though the example itself ignored the return code. Fixed at the library level (not just the example) to match `crate::ibm::quantum_compute_service::QuantumComputeService::task_stop`'s existing pattern: fetch the job's current status first, and only call `cancel_job_v1` if it's still `Waiting` or `Processing`. Unlike that existing implementation, this one does *not* discard the cancel call's own error: having just confirmed the job was cancellable, a failure at  that point is more likely a genuine problem (auth, network, a server error) than the job finishing in the split second between the check and the call, and swallowing a real failure would leave the caller believing the job was stopped when it wasn't.

### Backward compatibility

None of this has shipped yet, so return codes were renumbered/retired directly rather than leaving gaps for compatibility (109 is reused for the merged code; 110 is simply unused now). No other `QrmiError`/`ReturnCode`/
exception outside what's listed above changed.

### Testing

- `cargo build --release`
- `maturin build --release`
- Manual testing against a live IQM Server backend: invalid API token → 401
  → `AuthenticationFailed`; malformed job payload → 400 → `InvalidInput`;
  unknown job ID → 404 → `TaskNotFound`.

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2581,6 +2581,7 @@ dependencies = [
  "ffi_helpers",
  "futures",
  "glob",
+ "http 1.4.0",
  "iqm-server-api",
  "libc",
  "log",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ munge = ["pasqal-local-api/munge"]
 
 [dependencies]
 reqwest-retry = "0.7.0"
+http = "1.1.0"
 tokio = { version = "1.40.0", features = ["full"] }
 env_logger = "0.11.5"
 serde = { version = "1.0", features = ["derive"] }
--- a/dependencies/iqm_client/src/models/iqm_server_qc_health_status.rs
+++ b/dependencies/iqm_client/src/models/iqm_server_qc_health_status.rs
@@ -39,10 +39,7 @@ pub struct IqmServerQcHealthStatus {
 
 impl IqmServerQcHealthStatus {
     /// The current computer health status
-    pub fn new(
-        operational: String,
-        health: models::QcHealthDetail,
-    ) -> IqmServerQcHealthStatus {
+    pub fn new(operational: String, health: models::QcHealthDetail) -> IqmServerQcHealthStatus {
         IqmServerQcHealthStatus {
             operational,
             health: Box::new(health),
@@ -64,10 +61,7 @@ pub struct QcHealthDetail {
 }
 
 impl QcHealthDetail {
-    pub fn new(
-        healthy: bool,
-        updated_at: chrono::DateTime<chrono::FixedOffset>,
-    ) -> QcHealthDetail {
+    pub fn new(healthy: bool, updated_at: chrono::DateTime<chrono::FixedOffset>) -> QcHealthDetail {
         QcHealthDetail {
             healthy,
             updated_at,
--- a/dependencies/iqm_client/src/models/iqm_server_qc_health_status.rs
+++ b/dependencies/iqm_client/src/models/iqm_server_qc_health_status.rs
@@ -15,9 +15,46 @@
 use crate::models;
 use serde::{Deserialize, Serialize};
 
-/// IqmServerQcHealthStatus : The current computer health status
+/// IqmServerQcHealthStatus : The current computer health status.
+///
+/// NOTE: hand-patched, not regenerated from the OpenAPI spec. The live IQM
+/// Server API wraps the health details one level deeper than what this
+/// crate was originally generated against --
+/// `{"operational": "online", "health": {"healthy": true, "updated_at":
+/// "..."}}` -- instead of the flat `{"healthy": ..., "updated_at": ...}`
+/// this struct used to be. `get_qc_health_v1` (in
+/// `src/apis/quantum_computers_api.rs`) deserializes the whole response
+/// body directly into this type, unchanged, so fixing the shape here was
+/// enough without touching that function. Revert this patch once the crate
+/// is regenerated from an OpenAPI spec that reflects the server's current
+/// response shape.
 #[derive(Clone, Default, Debug, PartialEq, Serialize, Deserialize)]
 pub struct IqmServerQcHealthStatus {
+    /// Whether the quantum computer is currently accepting jobs.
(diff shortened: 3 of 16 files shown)


yoonho (approved):
LGTM. Thanks for taking care this so quickly.

PR #214, opened 2026-08-24
[IBM] Classify Qiskit Runtime API errors (Qiskit Runtime Service and Quantum Compute Service)

ohtanim:
## Description of Change

@OkuyanBoga @oscar-wallis @m-meller @jcornall STFC team,
Could you please review and approve this PR at your earliest convenience? This PR contains one of the recent typed error support changes. The implementation is relatively simple and only introduces a mapping from anyhow::Result to QrmiError. I am hoping to release this change as soon as possible. If you do not expect to have time to review it in the near future, please let me know. I will then explore alternative review arrangements, both for this PR and for future code reviews.

### Summary

Extends the typed-error classification work done for `quantum_system_client` (#212) and IQM (#213) to `quantum_compute_client` -- the OpenAPI-generated client shared by both `IBMQiskitRuntimeService` and
`IBMQuantumComputeService`. Both implementations wrapped every call with `anyhow::Context`, so every failure surfaced as `QrmiError::Other` regardless of cause; a 401 looked the same as a 404 or a network timeout.

#### Shared classifier (`src/ibm/error.rs`)
- `quantum_compute_client` is OpenAPI-generated code we don't own or modify (same constraint as `iqm_server_api`), and -- unlike IQM -- every operation here reuses the *same* generic error body model
  (`ListJobs400Response`, just a `trace`/`errors` pair) regardless of status, rather than a distinct named model per status. There's nothing in the type system to confirm what a given status actually means beyond the bare code itself, so `classify::<T>()` works off `Error::ResponseError`'s `status` field the same way IQM's classifier does, but is more
  conservative about which statuses it's willing to name: 
  - 400 → `InvalidInput` (the reused model's own name suggests it was originally modeled for exactly this).
  - 401 → `AuthenticationFailed` (consistent across every endpoint). 
  - 404 → `ResourceNotFound`/`TaskNotFound`, disambiguated by a `ResourceKind` parameter (`Backend`/`Job`/`Session`) passed at each call site -- the status alone doesn't say which kind of thing was missing.
  - A `Session` 404 is deliberately left unclassified rather than forced into `ResourceNotFound`: a session isn't a compute resource the way a backend is, and `QrmiError` doesn't have a variant for "this session/acquisition doesn't exist," so mislabeling it seemed worse than leaving it generic.
  - 403 shows up on almost every endpoint here, right alongside 401 -- but with no distinguishing model, and IQM having already taught us not to assume 403 means "access denied" without real evidence, it's deliberately left unclassified too.
  - Everything unclassified (403, a `Session` 404, 409, 429, 500, ...) falls through to `QrmiError::Other`, but with the response body attached via `.context(...)` -- `Error<T>`'s own `Display` only shows the bare status (e.g. `"status code 403 Forbidden"`), dropping whatever the server actually said. `err` itself is kept as `source` either way, so nothing is lost for anyone inspecting the chain.

#### `src/ibm/qiskit_runtime_service.rs` / `src/ibm/quantum_compute_service.rs`
- Both files are near-identical (same underlying client, same operations), so the same 12 `.context(...)?`/`.with_context(...)?` call sites in each were replaced with `.map_err(|e| classify(e, ResourceKind::...))?`.
- No behavior changes beyond the error classification itself.

### Backward compatibility

Purely additive from a behavior standpoint: calls that previously succeeded still succeed the same way. Calls that previously failed still fail, just with a more specific `QrmiError` variant (and correspondingly more specific
`ReturnCode`/Python exception) instead of always `Other`. No new `QrmiError`/`ReturnCode`/exception variants were needed -- everything maps onto ones already introduced by the earlier IQM work #213 .

### Testing

- `cargo build --release`
- `maturin build --release`
- Manual testing against a live IBM Qiskit Runtime backend: 

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/src/ibm/qiskit_runtime_service.rs
+++ b/src/ibm/qiskit_runtime_service.rs
@@ -12,7 +12,7 @@
 
 // This code is part of Qiskit.
 //
-// (C) Copyright IBM 2025
+// (C) Copyright IBM 2025-2026
 //
 // This code is licensed under the Apache License, Version 2.0. You may
 // obtain a copy of this license in the LICENSE.txt file in the root directory
--- a/src/ibm/qiskit_runtime_service.rs
+++ b/src/ibm/qiskit_runtime_service.rs
@@ -23,13 +23,12 @@
 // that they have been altered from the originals.
 
 use crate::error::{required_env, QrmiError};
-use crate::ibm::error::IbmError;
+use crate::ibm::error::{classify, IbmError, ResourceKind};
 use crate::ibm::qiskit_runtime_service::models::{
     CreateJobRequestOneOfAllOfParams, EstimatorV2Input, NoiseLearnerInput, SamplerV2Input,
 };
 use crate::models::{Payload, ResourceType, Target, TaskResult, TaskStatus};
 use crate::{QuantumResource, Result};
-use anyhow::Context;
 use log::error;
 use quantum_compute_client::apis::{auth, backends_api, configuration, jobs_api, sessions_api};
 use quantum_compute_client::models;
@@ -138,7 +137,7 @@ impl QuantumResource for IBMQiskitRuntimeService {
         let status_response =
             backends_api::get_backend_status(&self.config, &self.backend_name, None)
                 .await
-                .context("failed to get backend status")?;
+                .map_err(|e| classify(e, ResourceKind::Backend))?;
         // Print the status, using "unknown" if no status is available
         let status_str = status_response
             .status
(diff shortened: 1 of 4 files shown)


src/ibm/qiskit_runtime_service.rs
m-meller: Please add 2026 to copyright

m-meller (changes requested):
Thank you @ohtanim LGTM!
The only comment I have is to update copyright in `qiskit_runtime_service.rs`

src/ibm/qiskit_runtime_service.rs
ohtanim: Thanks. Fixed.

m-meller (approved):
Good to go! Thank you @ohtanim

PR #215, opened 2026-08-24
[AnB] Classify Alice & Bob Felis API errors, fix several panics on bad input

ohtanim:
## Description of Change

@machjn Could you please review and approve this PR at your earliest convenience? This PR contains one of the recent typed error support changes. The implementation is relatively simple and only introduces a mapping from anyhow::Result to QrmiError. I am hoping to release this change as soon as possible. If you do not expect to have time to review it in the near future, please let me know. 

I have not yet tested this in a production environment. While I have confirmed that a properly typed error is returned when the required environment variables are missing, I would appreciate it if you could help validate the implementation on an actual system. In particular, it would be helpful to verify how it behaves in cases such as authentication failures, missing resources, and missing tasks. Thanks,

@awennersteen @OkuyanBoga You don't need to review this PR.

### Summary

Extends the typed-error classification work done for `quantum_system_client` (#212), IQM (#213 ), and `quantum_compute_client` (#214 ) to `alice_bob_felis` -- the OpenAPI-generated client used by `AliceBobFelis`.
Also fixes four places that could panic on ordinary bad input (a malformed API key, invalid JSON in a job payload, an unmatched target, and an unexpected empty response) rather than returning a proper `QrmiError`.

#### Error classification (`src/alice_bob/error.rs`, new)
- Same general approach as [`crate::ibm::error`]'s classifier for `quantum_compute_client` (#214 ) : `alice_bob_felis` is OpenAPI-generated code we don't own or modify, so `classify::<T>()` works off `Error::ResponseError`'s `status` field rather than anything specific to this crate's generated code.
- This API's spec types even less than `quantum_compute_client`'s did: every operation this crate uses only ever types a single status -- 422 (`HttpValidationError`, FastAPI's standard validation-failure response) -- with nothing else having a typed model at all, generic or otherwise. That's different from IQM's 403 case, where a specific status had a *documented, differently-named* model proving it meant something else entirely; here there's simply no documentation either way for 401/404. Rather than assume the API deviates from ordinary HTTP semantics without evidence, 401 and 404 are still classified the standard way. 403 gets the same conservative treatment as everywhere else in this line of work, though: with nothing to confirm what it means here, it's left unclassified.
- Maps: 400/422 → `InvalidInput`; 401 → `AuthenticationFailed`; 404 → `ResourceNotFound`/`TaskNotFound`, disambiguated by a `ResourceKind` parameter (`Backend`/`Job`) passed at each call site. Everything else (403 included) falls through to `QrmiError::Other`, with the response body attached via `.context(...)` since `Error<T>`'s own `Display` only shows the bare status.
- `src/alice_bob/felis.rs`: all 7 `.context(...)?`/`.with_context(...)?` call sites replaced with `.map_err(|e| classify(e, ResourceKind::...))?`.

#### Panics on bad input (`src/alice_bob/felis.rs`)
Found while reviewing the file for this change:
- `AliceBobFelis::new()`: a malformed `..._QRMI_AB_FELIS_API_KEY` crashed the process via `decode_api_key(...).unwrap()`, even though `new()` already returns `Result<Self>`. Now returns `QrmiError::InvalidInput`.
- `task_start()`: malformed JSON in the job payload crashed the process via `serde_json::from_str(...).unwrap()`. Replaced with a bare `?`, which goes through the existing `QrmiError::InvalidJson` (`#[from]
  serde_json::Error`) -- no new variant needed.
- `target()`: no target matching the configured backend name crashed the process via `.expect(...)`. Now returns `QrmiError::ResourceNotFound`.
- `most_recent_event()`: an empty `events` list on a job crashed the process via `.unwrap()` on a comment-only assumption ("can safely assume events is non-empty"). Now returns `QrmiError::Other` instead of trusting
  that assumption unconditionally -- similar in spirit to the `get_qc_health_v1` response-shape issue found during the IQM work (#213 ):  trusting a vendor response to always have a certain shape without actually checking.

One remaining `.unwrap()` (`serde_json::to_string(&target).unwrap()` in `target()`) was left as-is: it's serializing our own already-typed internal data back to JSON, not anything a caller or vendor response controls, so it's not in the same category as the four above.

### Backward compatibility

Purely additive from a behavior standpoint: calls that previously succeeded still succeed the same way. Calls that previously failed with a generic `Other` now fail with a more specific `QrmiError` variant where classifiable. The four panic fixes change crash-the-process behavior into a normal `Err` return -- a behavior change, but one that only affects callers who were previously getting a panic (i.e., there's no "successful" behavior being altered). No new QrmiError`/`ReturnCode`/exception variants were needed -- everything maps onto ones already introduced by the earlier IQM work.

### Testing

- `cargo build --release`
- `maturin build --release`

## Checklist ✅

- [ ] Have you included a description of this change?

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/examples/qrmi/c/alice_bob_felis/src/felis.c
+++ b/examples/qrmi/c/alice_bob_felis/src/felis.c
@@ -32,8 +32,9 @@ int main(int argc, char *argv[]) {
   QrmiQuantumResource *qrmi =
       qrmi_resource_new(argv[1], QRMI_RESOURCE_TYPE_ALICE_BOB_FELIS);
   if (!qrmi) {
-    const char* last_error = qrmi_get_last_error();
-    fprintf(stderr, "Failed to create QRMI for %s. %s\n", argv[1], last_error);
+    const char *last_error = qrmi_get_last_error();
+    fprintf(stderr, "Failed to create QRMI for %s. %s (%d)\n", argv[1],
+            last_error, qrmi_get_last_error_kind());
     qrmi_string_free((char *)last_error);
     return EXIT_FAILURE;
   }
@@ -45,8 +46,10 @@ int main(int argc, char *argv[]) {
     QrmiResourceType resource_type;
     rc = qrmi_resource_type(qrmi, &resource_type);
     if (rc == QRMI_RETURN_CODE_SUCCESS) {
-      const char *resource_type_str = qrmi_config_resource_type_to_str(resource_type);
-      fprintf(stdout, "Selected resource: id=%s type=%s\n", resource_id, resource_type_str);
+      const char *resource_type_str =
+          qrmi_config_resource_type_to_str(resource_type);
+      fprintf(stdout, "Selected resource: id=%s type=%s\n", resource_id,
+              resource_type_str);
     }
     qrmi_string_free(resource_id);
   }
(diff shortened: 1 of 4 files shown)


machjn:
LGTM

machjn (approved):

PR #224, opened 2026-08-27
Update license to Apache-2.0

1ucian0:
Changed license format from file reference to string. In this way, the ambiguity on the license version is removed.

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for more context.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Physics",
 ]
 dynamic = ["version"]
-license = {file = "LICENSE.txt"}
+license = "Apache-2.0"
 readme = "README.md"
 description = "Quantum Resource Management Interface(QRMI)"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -25,7 +25,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Physics",
 ]
 dynamic = ["version"]
-license = {file = "LICENSE.txt"}
+license = "Apache-2.0"
```
burgholzer, line 28: Because I just happened to see this, you can still keep the pointer to the file by using the standard `license-files`:

```suggestion
license = "Apache-2.0"
license-files = ["LICENSE.txt"]
```
See https://packaging.python.org/en/latest/guides/licensing-examples-and-user-scenarios/#advanced-example

ohtanim (approved):
Thanks. LGTM.

burgholzer (approved):

PR #225, opened 2026-08-28
Embed build/version info into QRMI library

ohtanim:
### Review & Approve

Urgency: Not urgent. That said, a quick review would give us a chance to get this into the current customer upgrade window, so faster turnaround is appreciated if feasible.

Target date: Hoping to complete until early next week.

### Background

This is a companion change to [spank-plugins PR #202](https://github.com/qiskit-community/spank-plugins/pull/202) .
While investigating a quantum-workload issue during a customer cluster upgrade, we found it was hard to tell which QRMI build was actually linked into the deployed `spank_qrmi.so` without rebuilding or tracing through job
logs. The spank-plugins PR embeds spank_qrmi's own version into a `.version_info` ELF section; this PR does the equivalent for QRMI itself, so both pieces of the puzzle are visible from the same deployed binary with a single `strings`/`readelf` command.

### Summary

Embeds QRMI's crate version (`CARGO_PKG_VERSION`) and the exact git commit it was built from into a dedicated `.version_info` ELF section, so it can be inspected offline without running any code:

```bash
$ strings libqrmi.so | grep QRMI_BUILD_VERSION
QRMI_BUILD_VERSION:0.24.0;QRMI_GIT_HASH:0dac1793b013

$ readelf -p .version_info libqrmi.so
String dump of section '.version_info':
  [    4f]  QRMI_BUILD_VERSION:0.24.0;QRMI_GIT_HASH:0dac1793b013
```

This coexists cleanly with spank_qrmi's own marker in the same section once linked together (distinct symbol names, no collision):

```bash
$ strings spank_qrmi.so | grep -E "SPANK_QRMI|QRMI_BUILD"
SPANK_QRMI_VERSION=0.11.0;SPANK_QRMI_GIT_HASH=0dac1793b013
QRMI_BUILD_VERSION:0.24.0;QRMI_GIT_HASH:0dac1793b013

$ readelf -p .version_info spank_qrmi.so
String dump of section '.version_info':
  [     0]  SPANK_QRMI_VERSION=0.11.0;SPANK_QRMI_GIT_HASH=3845dd911381
  [    3b]  QRMI_BUILD_VERSION:0.24.0;QRMI_GIT_HASH:7a9573703ac4
```

### Changes

- **`build.rs`**: resolves the exact git commit QRMI was built from via `git rev-parse --short=12 HEAD` and exposes it to the crate at compile time as `GIT_HASH` via `cargo:rustc-env=`. Falls back to `"unknown"` when
  built outside a git checkout (e.g. `-DQRMI_ROOT=` pointing at an extracted tarball with no `.git` directory).
- **`lib.rs`**: adds a `VERSION_INFO` byte array built from `env!("CARGO_PKG_VERSION")` and `env!("GIT_HASH")`, placed into a `.version_info` ELF section via `#[link_section]`. A few details worth
  calling out explicitly in review, since they were each needed to make this actually show up in the final linked binary:
  - The string is assembled at **compile time** via `concat!`/`env!`/a `const fn` (compile-time function evaluation), not at runtime — there is no runtime cost and nothing "executes" when the plugin loads.
  - Uses a fixed-size `[u8; N]` array rather than `&[u8]`/`&str`: a slice reference is a fat pointer (pointer + length) and would put a relocatable pointer into `.version_info` instead of the actual text.
  - Requires **both** `#[used]` (survive Rust/LLVM dead-code elimination) **and** `#[no_mangle]` (export the symbol so it survives the linker's `--gc-sections`, and so the containing object file isn't dropped when QRMI is linked in as a static archive and only some of its other symbols are referenced).
- **`cbindgen.toml`**: adds `VERSION_INFO` to `[export] exclude`. Without this, cbindgen emits `extern const uint8_t VERSION_INFO[VERSION_LEN];` into the generated `qrmi.h`, but `VERSION_LEN` is a private (non-`pub`)
  const that cbindgen silently skips — leaving `VERSION_LEN` undeclared in the header and breaking the C build
  (`'VERSION_LEN' undeclared here`). `VERSION_INFO` is purely an offline-inspection marker, not part of the public C API, so it has no business being in the generated header at all.
  
### Testing

- Verified `readelf -p .version_info` / `strings` show the expected string on a locally built `cdylib` and after linking a `staticlib` build into a separate final binary (mirroring how spank_qrmi links against `libqrmi.a`), including with only an unrelated QRMI function referenced from the C side (confirms the object file isn't dropped during archive member pruning).
- Verified `cbindgen.toml` change: regenerated `qrmi.h` and confirmed no `VERSION_INFO`/`VERSION_LEN` declarations appear, and that `spank_qrmi.c` (from the companion PR) compiles cleanly against it.
- Confirmed `git rev-parse --short=12 HEAD` in `build.rs` matches the same 12-character abbreviation used on the spank_qrmi side, so the two hashes are directly comparable when both are visible in the same binary.
- full build via the actual `ExternalProject`/CMake fetch path used by spank-plugins (both default GitHub fetch and
  `-DQRMI_ROOT=` local-checkout path).

### Notes / follow-ups

- Not a behavior change to QRMI's actual functionality — purely additive metadata.

## Checklist ✅

   - Please ignore lychee error because spank plugin PR is still under review.

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -128,13 +128,13 @@ QRMI_BUILD_VERSION:0.24.0;QRMI_GIT_HASH:0dac1793b013
 ```
 
 The two are independent: `SPANK_QRMI_*` describes the plugin binary
-itself, `QRMI_BUILD_VERSION`/`GIT_HASH` describes the QRMI crate it was
-linked against. See the spank_qrmi plugin's own FAQ for details on the
+itself, `QRMI_BUILD_VERSION`/`QRMI_GIT_HASH` describes the QRMI crate it was
+linked against. See the spank_qrmi plugin's own [FAQ](https://github.com/qiskit-community/spank-plugins/blob/main/docs/FAQ.md) for details on the
 former.
 
 #### Notes
 
-- If `GIT_HASH` shows `unknown`, QRMI was most likely built from a source
+- If `QRMI_GIT_HASH` shows `unknown`, QRMI was most likely built from a source
   tree without a `.git` directory (e.g. an extracted release tarball, or a
   local checkout pointed at via `QRMI_ROOT` in the consuming project's
   build).
--- a/FAQ.md
+++ b/FAQ.md
@@ -87,14 +87,14 @@ any binary that links it), so you can check it without running the code.
 #### Using `strings`
 
 ```shell-session
-$ strings /path/to/spank_qrmi.so | grep QRMI_BUILD_VERSION
+$ strings /path/to/libqrmi.so | grep QRMI_BUILD_VERSION
 QRMI_BUILD_VERSION:0.24.0;QRMI_GIT_HASH:0dac1793b013
 ```
 
 #### Using `readelf`
 
 ```shell-session
-$ readelf -p .version_info /path/to/spank_qrmi.so
+$ readelf -p .version_info /path/to/libqrmi.so
 
 String dump of section '.version_info':
   [    4f]  QRMI_BUILD_VERSION:0.24.0;QRMI_GIT_HASH:0dac1793b013
--- a/FAQ.md
+++ b/FAQ.md
@@ -5,6 +5,8 @@
 1. [General Question](#general-questions)
 2. [Job Execution Errors](#job-execution-errors)
     1. [I get an error `error: spank_qrmi_c, failed to acquire resource: ibm_brisbane`](#i-get-an-error-error-spank_qrmi_c-failed-to-acquire-resource-ibm_brisbane)
+3. [Release and Deployment](#release-and-deployment)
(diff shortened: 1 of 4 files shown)


yoonho (approved):
LGTM. Definitely would have helped with the RPI debugging. Nice FAQ additions too.

ohtanim:
> LGTM. Definitely would have helped with the RPI debugging. Nice FAQ additions too.

@yoonho Oops. I found minor issue in FAQ.md and fixed. Could you please signoff again?

yoonho (approved):
LGTM

ohtanim:
> LGTM

Thanks for prompt action. Let us release this to our customer.

PR #172, opened 2026-07-28
[STFC] Add Python Test Coverage and Primitives Unit Tests

jcornall:
## Description of Change

This PR introduces a new CI workflow that runs `pytest` unit tests and reports on test coverage across the Python package. Additionally, this PR adds more unit tests to the Python package, aimed at testing the primitives modules. Copilot LLM was used to help write some of these tests and suggest fixes where they arose.

The workflow reports test coverage after running the below job - `--cov-fail-under=70` controls the threshold for coverage % tolerance:

```
      - name: Run unit tests
        continue-on-error: true
        run: |
          pytest \
            --cov=qrmi \
            --cov-report=term-missing \
            --cov-report=xml \
            --cov-fail-under=70
```

Additional jobs are included to post/update a comment on the PR reporting on the coverage, as well as printing the coverage % to the workflow summary.

The PR also modifies the `sphinx.yml` workflow to avoid running the action when only Python tests are being modified.

## Checklist ✅

- [x] Have you included a description of this change?
- [x] Have you updated the relevant documentation to reflect this change?
- [x] Have you made sure CI is passing before requesting a review?

## Ticket
- [x] Is Part Of #171 
- [x] Fixes #183 
- [x] Fixes #76

Issue #171, opened 2026-07-28
Add Additional Python Unit Tests


jcornall:
### What problem are you trying to solve?

The `/python/tests` directory currently contains two test modules. A quick test coverage report (using `pytest --cov=python --cov-report=term-missing`) highlights that many Python modules have 0% test coverage. Additional Python unit tests should be added to make the Python bindings more robust and flag any errors caused by changes to the codebase. 

### Describe the solution you’d like

Additional unit tests should be added to the `/python/tests/` folder. This issue will not add integration tests, which will instead be the focus of a separate issue.

PR #227, opened 2026-08-28
[STFC] Add Python Miscellaneous Unit Tests 

### LINKED ISSUE #184: Add Python Miscellaneous Unit Tests
QRMI's Python bindings stored within `/python/qrmi/` require unit tests to ensure that they are robust and operate correctly. This issue accounts for modules outside of the `/primitives/` directory.

The following modules require unit tests:

- `/pulser/service.py`
- `/pulser/connection.py`
- `/qiskit_iqm/iqm_provider.py`
- `/tools/task_runner/main.py`
- `__init__.py`

jcornall:
## Description of Change

This PR adds additional non-primitive unit tests to the Python package. Copilot LLM was used to help write some of these tests and suggest fixes where they arose. See #172 for the QRMI primitives unit tests.

## Checklist ✅

## Ticket
- [x] Is Part Of #171 
- [x] Fixes #184


pyproject.toml
```diff
@@ -48,15 +48,18 @@ dependencies = [
 
 [project.optional-dependencies]
 ibm = [
-  "qiskit_ibm_runtime>=0.30.0",
+  # 0.46+ requires Qiskit/Pydantic versions incompatible with IQM Client 35.x.
+  "qiskit_ibm_runtime>=0.30.0,<0.46.0",
```
ohtanim, line 52: This pyproject.toml is used when creating the QRMI PyPI package. I cannot agree with this change, as it means IBM Quantum users will no longer be able to use the latest version of Qiskit. As noted in the README, IQM and IBM packages cannot currently be installed in the same environment. Users would need to create separate Python environments for IBM and IQM and install the respective packages in each environment.

Please consider alternative approaches that would avoid changing pyproject.toml.

ohtanim (changes requested):

pyproject.toml
jcornall: Hi Ohtani-san, thank you for clarifying. This change was a workaround to get the unit tests to pass in the same environment, if IQM and IBM can't be installed together then I will look at ways of testing these separately in the CI/CD workflow.

ohtanim: Or you can create separated pyenv for CI testing.

jcornall: I've moved the test file these dependency issues affect to its own directory (`/python/tests/compatibility/`), the CI/CD workflow now runs these tests separately using the compatible versions - running these tests with the latest versions results in an error, so I have separated it from the standard testing suite. 

These tests are ignored in the wheels CI/CD workflow, as that should test core functionality.

oscar-wallis:
Hi @badtst , some of the files changed are pasqals, could we get your eyes on this for a review? Thanks!

python/qrmi/pulser/connection.py
```diff
@@ -140,9 +140,9 @@ def get_batch_logs(self, batch_id: str | None = None) -> Sequence[str]:
         for job_id in job_ids:
             try:
                 job_logs = self._qrmi.task_logs(job_id)
+                logs.append(job_logs)
```
ohtanim, line 143: @awennersteen @MatthieuMoreau0 @badtst Could you please verify this on your side? This is part of your PyPL code.

ohtanim (approved):

python/qrmi/pulser/connection.py
badtst: Ok with me, good catch on your side @jcornall

ohtanim: @badtst Can you walk thru and approve?

jcornall: Just copying here what I sent to Oscar for future reference:

I was writing a test for the except section in `get_batch_logs()` and it seemed like the line should be there instead of after the try/except block - if the try failed then `logs.append(job_logs)` was being called even though `job_logs` hadn't changed/been defined.

badtst (approved):
Ok with me

ohtanim:
let us merge this to main.

PR #228, opened 2026-08-29
Restrict .version_info embedding to Linux to fix macOS build failure

ohtanim:
## Review & Approve

**Urgency:** Not urgent, but this is currently blocking local builds on macOS, so a relatively quick look would help unblock anyone developing on a Mac.

**Target date:** within a week

### Summary

Building on macOS currently fails:

```
rustc-LLVM ERROR: Global variable 'VERSION_INFO' has an invalid section specifier '.version_info': mach-o section specifier requires a segment and section separated by a comma.
```

`.version_info` (no comma) is ELF-specific section-naming syntax. Mach-O (macOS) requires `"SEGMENT,section"` syntax instead (e.g. `"__TEXT,__version_info"`), and PE/COFF (Windows) has yet another set of constraints (8-character section name limit, no segment/section split).

Since this marker only matters for the actual Slurm/Linux deployment target, this PR wraps the whole `version_info` module (the string constant, the `const fn`, and the exported static) in `#[cfg(target_os = "linux")]` instead of trying to support every object format. On macOS/Windows the module is compiled out entirely, so local development on a Mac is no longer blocked by this.

### Changes

- **`lib.rs`**: wrapped the existing `VERSION_INFO` / `VERSION_STR` / `VERSION_LEN` / `str_to_array` items in a private
  `#[cfg(target_os = "linux")] mod version_info { ... }`. No change to the Linux behavior or the embedded string format.
- Also fixed some metadata names in FAQ.md.

### Testing

- Verified the Linux build still produces the expected `.version_info`  section (`readelf -p .version_info` / `strings` unchanged).
- Verified the MacOS build

### Notes / follow-ups

- Windows isn't handled either, for the same reason (incompatible section syntax) — not addressed here since it isn't currently blocking anyone, but the same `#[cfg(target_os = "...")]` pattern would extend cleanly if needed later.
- Not a behavior change on Linux — purely a build-fix for other platforms.

## Checklist ✅

- [ ] Have you included a description of this change?

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -80,6 +80,9 @@ You are not authorized to run a session when using the open plan.
 
 ### How can I check which version of QRMI is linked into a binary?
 
+> [!NOTE]
+> Linux only.
+
 Every build of QRMI embeds its own crate version and git commit hash
 directly into the compiled artifact (shared library, static library, or
 any binary that links it), so you can check it without running the code.
--- a/FAQ.md
+++ b/FAQ.md
@@ -122,13 +122,13 @@ rebuilding or adding logging.
 the same `.version_info` section:
 
 ```shell-session
-$ strings spank_qrmi.so | grep -E "SPANK_QRMI|QRMI_BUILD"
-SPANK_QRMI_VERSION=0.11.0;SPANK_QRMI_GIT_HASH=0dac1793b013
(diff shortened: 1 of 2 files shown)


yoonho (approved):
LGTM

PR #232, opened 2026-08-30
Add asyncio-native  Python binding

ohtanim:
## Description of Change

(Editted) To reviewers, 

This feature was originally added for a new QRMI vendor implementation I was preparing. However, after further discussions, we no longer have an immediate use case for it. If you still see value in this feature, please review and approve the PR. Otherwise, feel free to close it

### Summary

Adds an asyncio-native counterpart to the existing `QuantumResource` Python binding, built on `pyo3-async-runtimes`. Every I/O method (`is_accessible`, `acquire`, `release`, `task_start`, `task_status`, `task_result`, `task_logs`, `task_stop`, `target`, `metadata`, `resource_id`, `resource_type`) returns an `asyncio.Future` instead of blocking the calling thread, so it can be `await`ed from an asyncio event loop and multiple resources can be driven concurrently with `asyncio.gather`.

```python
import asyncio
from qrmi import AsyncQuantumResource, ResourceType

async def main():
    r = AsyncQuantumResource("ibm_torino", ResourceType.IBMQuantumSystem)
    if await r.is_accessible():
        lock = await r.acquire()
        ...

asyncio.run(main())
```

### Motivation

The existing `QuantumResource` binding blocks the calling thread for every I/O call (it drives its own per-instance tokio `Runtime` via `py.detach()`+`block_on()`). That's fine for a single script that submits one job and waits, but it has two real gaps:

- **Embedding in an asyncio app.** Anything already built on asyncio (an aiohttp/FastAPI service orchestrating QPU jobs, for instance) has to wrap every QRMI call in `loop.run_in_executor(...)` to avoid blocking its event loop. That's boilerplate that's easy to forget at one call site and get a stalled event loop for it.
- **Integrating with vendor SDKs that are themselves asyncio-based.** Some quantum vendor SDKs are implemented on top of asyncio. Mixing those with a blocking QRMI call means either wrapping QRMI in `run_in_executor` or wrapping the vendor SDK in its own thread — either way, two different concurrency models bolted together in the same program. An asyncio-native QRMI binding lets both sides `await` on the same event loop directly.
- **Concurrency across multiple resources.** Running the same workload against several backends at once currently means reaching for `threading`/`concurrent.futures` alongside the sync binding. `asyncio.gather` over several of these resources works today.

To be clear about what this *isn't*: it isn't a performance optimization for the common single-backend, single-job case, and by itself it doesn't make any individual QRMI call faster. Two different `AsyncQuantumResource` instances do run truly concurrently (see Design notes), but calls on the *same* instance still serialize through the underlying resource, exactly
as with the sync binding — this is about *not blocking the caller* and *composing cleanly with asyncio*, not about parallelizing work that was sequential before.

## What's added

- **`src/pyext_async.rs`** (new file) — the `AsyncQuantumResource` pyclass. Wraps the resource in `Arc<tokio::sync::Mutex<Box<dyn QuantumResource + Send + Sync>>>` so each method can hand `pyo3_async_runtimes`'s `future_into_py` a `'static` future without needing `&mut self` to outlive the call. Doesn't own a per-instance tokio `Runtime` like the sync binding does; it schedules work on the single process-wide runtime
  `pyo3-async-runtimes` lazily creates.
- **`src/pyext/common.rs`** (new file) — the `QrmiError` -> `PyErr`  exception hierarchy, `to_py_err`, and the `ResourceType` conversions (both directions), pulled out of `pyext.rs` so `pyext_async.rs` can reuse them via `pub(crate)` instead of duplicating them. Also newly adds a proper `impl From<crate::models::ResourceType> for ResourceType` for
  the reverse direction, which previously had to be hand-inlined at each of its two call sites in `pyext.rs`.
- **New Cargo feature `pyo3-async`** (`Cargo.toml`), separate from the existing `pyo3` feature, so existing consumers of the sync API aren't forced to pull in `pyo3-async-runtimes` + a global tokio runtime.
- **`src/lib.rs`**: `#[cfg(feature = "pyo3-async")] pub mod pyext_async;`
- **`src/pyext.rs`**: refactored, not just appended to. The exception hierarchy, `to_py_err`, and `ResourceType`'s definition moved out into `pyext::common` (see above); the two remaining call sites
  (`PyQuantumResource::resource_type` and `PyResourceDef::resource_type`) now call `ResourceType::from(...)` instead of repeating the match by hand. The one line actually *added* (not moved) is the new class registration in the existing `#[pymodule] fn qrmi(...)`.
- **Examples** (`examples/qrmi/python/async/`):
  - `async_example.py` — asyncio port of the existing IBM Quantum System example
  - `is_accessible_concurrent.py` — checks `is_accessible()` on multiple backends concurrently via `asyncio.gather`
  - `task_run_concurrent.py` — submits a task to multiple backends concurrently (one input file + one program id per backend, paired via repeated `--backend`/`--input`/`--program-id` flags), polls each independently, and collects all results (or per-backend errors, via `return_exceptions=True`)

### Design notes

- Two different `AsyncQuantumResource` instances run truly concurrently (independent `Arc<Mutex<..>>`, shared multi-threaded tokio runtime); concurrent calls on the *same* instance still serialize through its internal mutex, matching the existing single-instance semantics of the sync binding.
- No changes to the `QuantumResource` trait, `models`, or any existing provider implementation (IBM/Pasqal/IQM/Alice&Bob) — this is purely an additive binding layer plus a same-crate refactor to avoid duplicating
  error/type conversion logic across the two bindings.
  
### Testing

- `cargo build --release
- `maturin build --release`, installed into a venv, confirmed `AsyncQuantumResource` is importable and functionally
  equivalent to the sync API against a real IBM Quantum System backend (`is_accessible`, `acquire`/`release`, `task_start` → poll `task_status` → `task_result`/`task_logs` → `task_stop`)
- Ran `is_accessible_concurrent.py` and `task_run_concurrent.py` against multiple real backends concurrently
- Ran existing 'sync' python examples against real backends

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2484,6 +2484,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "pyo3-async-runtimes"
+version = "0.29.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b3ef68daa7316a3fac65e5e18b2203f010346de1c1c53456811a2624673ab046"
+dependencies = [
+ "futures-channel",
+ "futures-util",
+ "once_cell",
+ "pin-project-lite",
+ "pyo3",
+ "tokio",
+]
+
+[[package]]
 name = "pyo3-build-config"
 version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2588,6 +2602,7 @@ dependencies = [
  "pasqal-cloud-api",
  "pasqal-local-api",
  "pyo3",
+ "pyo3-async-runtimes",
  "pyo3-stub-gen",
  "quantum-system-api",
  "quantum_compute_client",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,10 @@ required-features = ["build-binary"]
 [features]
 default = ["quantum-system-api/iqp_retry_policy"]
 pyo3 = ["dep:pyo3", "dep:pyo3-stub-gen"]
+# asyncio-native `AsyncQuantumResource` bindings (src/pyext_async.rs).
+# Separate from `pyo3` so existing consumers aren't forced to pull in
+# pyo3-async-runtimes + a global tokio runtime just to build the sync API.
+pyo3-async = ["pyo3", "dep:pyo3-async-runtimes"]
 build-binary = ["dep:signal-hook-tokio", "dep:signal-hook", "dep:eyre"]
 munge = ["pasqal-local-api/munge"]
 
@@ -48,6 +52,7 @@ anyhow = { version = "1.0.97", features = ["backtrace"] }
 thiserror = "2.0"
 glob = "0.3"
 pyo3 = { version = "0.29.0", features = ["anyhow"], optional = true }
+pyo3-async-runtimes = { version = "0.29", features = ["tokio-runtime"], optional = true }
 ffi_helpers = "0.3.0"
 async-trait = "0.1.88"
 dotenv = "0.15.0"
--- a/examples/qrmi/python/async/async_example.py
+++ b/examples/qrmi/python/async/async_example.py
@@ -1,0 +1,71 @@
+#
+# (C) Copyright 2026 IBM. All Rights Reserved.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""An example of IBM Quantum System QRMI python-bindings (asyncio version)"""
+
+import asyncio
+import json
+import argparse
+from dotenv import load_dotenv
+from qrmi import AsyncQuantumResource, ResourceType, Payload, TaskStatus
+
+
+async def main(args: argparse.Namespace) -> None:
+    """main"""
+    qrmi = AsyncQuantumResource(args.backend, ResourceType.IBMQuantumSystem)
+    print(qrmi)
+    print(
+        f"Selected resource: id={await qrmi.resource_id()} "
+        f"type={str(await qrmi.resource_type())}"
+    )
+
+    print(await qrmi.is_accessible())
+
+    lock = await qrmi.acquire()
+    print(f"lock {lock}")
+
+    target_json = json.loads((await qrmi.target()).value)
+    print(json.dumps(target_json, indent=2))
+    print(await qrmi.metadata())
+
+    with open(args.input, encoding="utf-8") as f:
+        primitive_input = f.read()
+        payload = Payload.QiskitPrimitive(input=primitive_input, program_id=args.program_id)
+        job_id = await qrmi.task_start(payload)
+        print(f"Task started {job_id}")
+
+        while True:
+            status = await qrmi.task_status(job_id)
+            if status not in [TaskStatus.Running, TaskStatus.Queued]:
+                break
+
+            await asyncio.sleep(1)
+
+        print(f"Task ended - {await qrmi.task_status(job_id)}")
+        print((await qrmi.task_result(job_id)).value)
+
+        print(await qrmi.task_logs(job_id))
+
+        await qrmi.task_stop(job_id)
+
+    await qrmi.release(lock)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="An example of IBM Quantum System QRMI")
+    parser.add_argument("backend", help="backend name")
+    parser.add_argument("input", help="primitive input file")
+    parser.add_argument("program_id", help="'estimator' or 'sampler'")
+    parsed_args = parser.parse_args()
+
+    load_dotenv()
+
+    asyncio.run(main(parsed_args))
--- a/examples/qrmi/python/async/is_accessible_concurrent.py
+++ b/examples/qrmi/python/async/is_accessible_concurrent.py
@@ -1,0 +1,49 @@
+#
+# (C) Copyright 2026 IBM. All Rights Reserved.
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""Check accessibility of multiple IBM Quantum System backends concurrently."""
+
+import asyncio
+import argparse
+from dotenv import load_dotenv
+from qrmi import AsyncQuantumResource, ResourceType
+
+
+async def check_one(backend: str) -> tuple[str, bool]:
+    """Check accessibility of one backend"""
+    qrmi = AsyncQuantumResource(backend, ResourceType.IBMQuantumSystem)
+    accessible = await qrmi.is_accessible()
+    return backend, accessible
+
+
+async def main(backends: list[str]) -> None:
+    """main"""
+
+    # Each `check_one(...)` call creates its own AsyncQuantumResource, so
+    # the awaits below run concurrently on QRMI's shared tokio runtime —
+    # this takes roughly as long as the single slowest backend check, not
+    # the sum of all of them.
+    results = await asyncio.gather(*(check_one(b) for b in backends))
+
+    for backend, accessible in results:
+        print(f"{backend}: is_accessible={accessible}")
+
+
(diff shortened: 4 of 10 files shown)


awennersteen:
I'm not terribly familiar with asyncio. So I asked an LLM to look at it and this is the raw response.

I will investigate what it reported tomorrow. But given that this was open a few days and the timezone difference I prefer to send it to you raw. But I'm planning to look at this myself tomorrow as I'd like to understand.



```
## Review: request changes

The low-level Rust implementation is broadly reasonable. Using `pyo3-async-runtimes`, a shared Tokio runtime, and an `Arc<tokio::sync::Mutex<...>>` is a defensible way to bridge the existing asynchronous Rust trait into Python. The mutex is necessary under the current `QuantumResource` trait because its methods require `&mut self`, and serializing operations on one resource instance is consistent with the existing semantics.

However, I do not think the proposed public Python API should be merged as-is. It does not behave like a conventional Python `async def` API, it covers only part of QRMI’s public abstraction, and its cancellation and typing semantics are insufficiently defined.

### 1. These methods return eagerly scheduled Futures, not Python coroutines

Every method calls `pyo3_async_runtimes::tokio::future_into_py()` immediately. Despite the wording in the PR, this does not return a normal Python coroutine object. It creates an `asyncio.Future` and spawns the Rust future on the Tokio runtime before returning it.

This distinction has observable consequences:

* Calling `resource.task_start(payload)` may begin the remote operation even if the returned object is never awaited.
* `inspect.iscoroutinefunction(resource.task_start)` is false.
* `asyncio.gather(...)` works because it accepts arbitrary awaitables.
* `asyncio.create_task(resource.task_start(...))` does not work because `create_task()` requires a coroutine.
* `TaskGroup.create_task(resource.task_start(...))` similarly raises `TypeError: a coroutine was expected, got <Future ...>`.

This directly contradicts the PR’s claim that each method “returns a Python coroutine” and its stated support for composition with `asyncio.TaskGroup`.

Python deliberately distinguishes coroutines, Tasks, and Futures. Calling a normal coroutine does not schedule it. [`TaskGroup.create_task()` accepts a coroutine](https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup.create_task), while [`future_into_py()` creates a Future and immediately spawns the Rust future](https://github.com/PyO3/pyo3-async-runtimes/blob/v0.29.0/src/generic.rs#L608-L669).

I consider this the main conceptual problem with the current API.

### 2. Suggested public API: a Python `async def` facade

I suggest keeping the Rust bridge as a private implementation type, for example `_AsyncQuantumResource`, and exposing a small public Python facade:

```python
class AsyncQuantumResource:
    def __init__(
        self,
        resource_id: str,
        resource_type: ResourceType,
    ) -> None:
        self._inner = _AsyncQuantumResource(resource_id, resource_type)

    async def is_accessible(self) -> bool:
        return await self._inner.is_accessible()

    async def task_start(self, payload: Payload) -> str:
        return await self._inner.task_start(payload)

    async def task_status(self, task_id: str) -> TaskStatus:
        return await self._inner.task_status(task_id)
```

This gives the public methods normal Python coroutine semantics:

* Operations begin when the coroutine is awaited or scheduled, not when the method is called.
* The methods work naturally with `asyncio.create_task()` and `TaskGroup.create_task()`.
* `inspect.iscoroutinefunction()` behaves as expected.
* The public signatures can have accurate Python return types.
* The PyO3 implementation can change without changing the public API.
* Python-specific facilities such as context managers and cancellation handling can be added without complicating the Rust binding.

The extra Python call layer should be negligible relative to the network operations performed by QRMI.

### 3. The async API covers only the lowest-level resource class

The PR exposes only [`[AsyncQuantumResource](https://github.com/qiskit-community/qrmi/blob/4b5db35f697f9ca51ef42b0765b502d4b7d1ee42/src/pyext_async.rs#L42-L174)`](https://github.com/qiskit-community/qrmi/blob/4b5db35f697f9ca51ef42b0765b502d4b7d1ee42/src/pyext_async.rs#L42-L174). There is no asynchronous equivalent of:

* `QRMIService`
* `ResourceProvider`

The source even says that `from_inner()` is used by `PyAsyncQRMIService::create()`, but no such class exists.

This matters conceptually for QRMI. `QRMIService` is the high-level entry point that discovers the QPU resources assigned to the current scheduler job. The new examples instead construct named resources directly. That approach is valid for low-level examples, but it should not become the only route available to async users because it steers them away from the resource-manager-controlled path.

If asynchronous support is intended as a first-class public API, it should cover the complete abstraction:

```python
from qrmi.aio import QRMIService

service = await QRMIService.create()
resource = service.resource("ibm_torino")

task_id = await resource.task_start(payload)
```

A `qrmi.aio` namespace would also avoid adding an expanding family of `AsyncQuantumResource`, `AsyncQRMIService`, and `AsyncResourceProvider` names to the top-level package. A top-level `AsyncQuantumResource` is not inherently wrong, but the namespace approach scales better if the full API is provided.

### 4. Cancellation semantics need to be defined

`pyo3-async-runtimes` propagates cancellation of the Python Future to the Rust future. This is straightforward for read-only operations such as:

* `is_accessible()`
* `task_status()`
* `task_result()`
* `target()`

It is much less straightforward for state-changing operations:

* `acquire()`
* `release()`
* `task_start()`
* `task_stop()`

Cancellation of a local future cannot establish whether a remote operation was committed. For example, the backend may create a session during `acquire()`, while cancellation prevents the client from receiving and storing the acquisition token. The caller can then be unable to release that session. Similarly, cancelling `task_start()` does not establish that no task was submitted.

The PR currently presents cancellation propagation as an advantage without discussing this ambiguity. At minimum, the public documentation should state the guarantees for each state-changing method, and tests should cover cancellation before, during, and after the underlying request.

An async context manager would improve the normal acquisition lifecycle:

```python
async with resource.lease():
    task_id = await resource.task_start(payload)
```

This would ensure that release is attempted after a successfully completed acquisition. It does not fully solve cancellation during the acquisition RPC, which may require idempotency or reconciliation support from the underlying provider.

### 5. Stub generation does not currently include the async API

The wheel enables `pyo3-async`, but the stub-generation, lint, and test targets still compile the stub generator with only `--features="pyo3"`. For example, [`[build-stubgen](https://github.com/qiskit-community/qrmi/blob/4b5db35f697f9ca51ef42b0765b502d4b7d1ee42/Makefile#L29-L33)`](https://github.com/qiskit-community/qrmi/blob/4b5db35f697f9ca51ef42b0765b502d4b7d1ee42/Makefile#L29-L33) does not enable `pyo3-async`.

Consequently, `AsyncQuantumResource` will not be included in the generated stubs.

Even if that feature is enabled, the Rust methods currently return `Bound<PyAny>`. `pyo3-stub-gen` maps this to `typing.Any`, so users would still not see that, for example:

```python
await resource.is_accessible()
```

returns `bool`, or that:

```python
await resource.task_status(task_id)
```

returns `TaskStatus`.

A Python facade would solve this naturally. Otherwise, every method needs an explicit stub return-type override such as `Awaitable[bool]`, `Awaitable[str]`, or `Awaitable[TaskStatus]`.

### 6. The “same event loop” description is inaccurate

The PR says that QRMI and asyncio-based vendor SDKs can await on the same event loop directly. That is not how `pyo3-async-runtimes` operates.

Python coroutines run on the Python `asyncio` loop. The QRMI Rust futures run on a separate, process-wide Tokio runtime. The library bridges completion and cancellation between those two runtimes. Its own documentation explicitly explains that [[Python and Rust futures use separate event loops](https://docs.rs/pyo3-async-runtimes/latest/pyo3_async_runtimes/#why-two-event-loops)](https://docs.rs/pyo3-async-runtimes/latest/pyo3_async_runtimes/#why-two-event-loops).

The user-facing result is asyncio-compatible and non-blocking, which is valuable, but the description should not imply that the Rust work runs on Python’s event loop.

### Recommendation

I agree that native asyncio support has value, particularly for services that coordinate several resources or integrate QRMI into FastAPI, aiohttp, or another asynchronous application.

However, I recommend one of the following:

1. Close or defer this PR until there is a concrete consumer that can validate the required public abstraction.

2. Request changes and introduce a complete, deliberately designed asynchronous API:

   * Keep the PyO3 Future-producing class private.
   * Add a public Python `async def` facade.
   * Include asynchronous `QRMIService` and `ResourceProvider` support.
   * Define cancellation semantics for state-changing operations.
   * Fix stub generation and return typing.
   * Add automated concurrency, lifecycle, and cancellation tests.

The underlying bridge is a useful implementation component. The concern is primarily that it is currently being exposed directly as QRMI’s long-term public asynchronous Python API before the semantics of that API have been settled.
```

ohtanim:
Thanks, AI. Good review. I will work on some points you mentioned above.

ohtanim:
@awennersteen I updated PR descriptions. It originally described about TaskGroup or coroutine but those are incorrect. I have updated with Future. Could you please ask your AI again to look into ? Regarding ResourceProvider and QRMIService, we don't need. ResourceProvider is early evaluation version, QRMIService only provides sync API.

PR #233, opened 2026-09-03
Add rpm libqrmi.spec as a reference

cclaudio:
## Description of Change

The QRMI project is not distributing RPM packages. However, STFC asked
if the community could provide a libqrmi.spec file so that they can
build the RPM locally. The libqrmi.spec covers only the libqrmi.so and
qrmi.h.

This PR adds the libqrmi.spec file, which builds two packages libqrmi
and libqrmi-devel. Development files (unversioned libqrmi.so symlink and
qrmi.h) goes in the libqrmi-devel package, which is a best practice
followed by rpm-based distros.

Recommended build instructions:

    $ ./run_in_container.sh make libqrmi-rpm

This PR also updates the CI to build the libqrmi rpms on a weekly basis

## Checklist ✅

## Ticket
- [ ] Fixes #
- [ ] Is Part of #


cclaudio:
CC: @oscar-wallis

ohtanim:
We [previously discussed RPM distribution in the Community](https://qiskit.slack.com/archives/C08ETNQDZK4/p1772491525823179?thread_ts=1772463917.284289&cid=C08ETNQDZK4). 

As I recall, the conclusion from that discussion was that RPM distribution doesn't match the use case for HPC datacenters. I'm confused as to why this PR is coming up now. Could you clarify once again what HPC use case actually requires RPM distribution? 

If this is something STFC-specific, I think it would be best for them to do the packaging themselves.

packaging/rpm/libqrmi.spec
```diff
@@ -0,0 +1,113 @@
...
+Version:        %{version}
+Release:        1%{?dist}
+Summary:        Quantum Resource Management Interface (QRMI) dynamic library
+License:        Apache-2.0
+URL:            https://github.com/qiskit-community/qrmi
+Source0:        qrmi-%{version}.tar.gz
+Source1:        qrmi-vendor-%{version}.tar.gz
+
+Provides:       libqrmi.so()(64bit)
+
```
ohtanim, line 30: I believe, at very least, we should add the following "Required" fields.

Requires: python3 >= 3.11
Requires: glibc >= 2.28

QRMI assumes python version should be 3.11 or above, and our ABI is built for glibc 2.28. However, I have no idea which RedHat version meets those requirements.

packaging/rpm/libqrmi.spec
```diff
@@ -0,0 +1,113 @@
...
+# The version is intentionally not hardcoded here. It must be supplied by the
+# caller via --define so that Cargo.toml remains the single source of truth.
+
+# Disable automatic debuginfo/debugsource package generation
+%global debug_package %{nil}
+
+Name:           libqrmi
+Version:        %{version}
+Release:        1%{?dist}
+Summary:        Quantum Resource Management Interface (QRMI) dynamic library
```
ohtanim: You're including qrmi.h file in addition to dynamic link library, correct? In this case, this Summary field is in-accurate.

Makefile
```diff
@@ -184,7 +184,8 @@ clean-c-examples:
 
 clean-tarballs:
 	rm -f $(DIST_DIR)/libqrmi-$(QRMI_VERSION)-el8-x86_64.tar.gz
-	rm -f $(DIST_DIR)/qrmi-$(QRMI_VERSION)-vendor.tar.gz
```
ohtanim, line 187: Is it okay to remove this line? We are already distributing the source under this file name. In other words, this amounts to a kind of external interface change. I would recommend to keep this line.

ohtanim (changes requested):

cclaudio:
> We [previously discussed RPM distribution in the Community](https://qiskit.slack.com/archives/C08ETNQDZK4/p1772491525823179?thread_ts=1772463917.284289&cid=C08ETNQDZK4).
> 
> As I recall, the conclusion from that discussion was that RPM distribution doesn't match the use case for HPC datacenters. I'm confused as to why this PR is coming up now. Could you clarify once again what HPC use case actually requires RPM distribution?
> 
> If this is something STFC-specific, I think it would be best for them to do the packaging themselves.

I know that STFC would like to build the libqrmi rpm themselves. They found the rpm spec file in the PR #68 and asked me if I am planning to submit an updated version of the rpm spec file. Maybe @oscar-wallis could shine in with some more details, if needed.

I agree with the concerns of distributing rpm/deb packages in the upstream project. I was just thinking if the same concerns apply to having the rpm spec file in the repository (as a reference for consumers) without distributing the actual rpm package. That's what motivated me to submit this PR

Thanks for the feedback, I can apply them after we decide if we want to go ahead with the PR.

ohtanim:
OK, let us wait for the clarifications from STFC. If they need this help, let’s consider.

I don’t have strong opinions on this.

awennersteen:
I'm ok either way.

An alternative will be to make a new repo in qrmi-commuity for packaging.
And we can toss in there rpm, deb, but also thinkgs like spack and easybuild

oscar-wallis:
Hi @cclaudio @ohtanim, our technical solutions group recently said their preferred installation method was via an rpm spec file, and they found Claudio's previous work but had some problems using it. This prompted me to ask if that work was still underway and the updated version here. Whether other datacentres would prefer this packaging method I couldn't tell you, but there is at least appetite at STFC, so access to it somewhere would be good.

packaging/rpm/libqrmi.spec
cclaudio: This rpm spec file covers only the libqrmi.so and qrmi.h.  Could you elaborate why python3 is required?

ohtanim: You're right. We don't need python3.

Makefile
cclaudio: I added the `QRMI_VENDOR_TARBALL_PATH` variable so the patch is actually replacing that by `rm -f $(QRMI_VENDOR_TARBALL_PATH)`, which expands to the same.

packaging/rpm/libqrmi.spec
cclaudio: I improved both the Summary and Description.

cclaudio:
Thanks @ohtanim and @awennersteen for the feedback. I addressed all of them and also made the following changes:
- Removed `Requires: zlib-devel` since it is not required
- Updated Summary and Description
- Modified the spec file to generate the libqrmi-devel package as well. It is a best practice to have header files and unversioned libs in a -devel package.
- Updated the weekly CI to also build the libqrmi rpms

.github/workflows/libqrmi.yml
```diff
@@ -254,6 +266,11 @@ jobs:
         path: libqrmi-${{ needs.info.outputs.qrmi-version }}*.tar.gz
         if-no-files-found: error
         retention-days: 1
+    - if: inputs.libqrmi-rpm == true
+      name: Build libqrmi rpms
+      run: |
+        dnf install -y rpm-build gcc make openssl-devel pkg-config
```
ohtanim: AFAIK, openssl-devel is no longer required. We elimimated the dependencies of openssl in #67.

Actually, ldd does not show any dependency on OpenSSL.

```
# ldd libqrmi.so 
linux-vdso.so.1 (0x0000ffffb014c000)
libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x0000ffffb00df000)
libm.so.6 => /lib64/libm.so.6 (0x0000ffffb003e000)
libc.so.6 => /lib64/libc.so.6 (0x0000ffffaee52000)
/lib/ld-linux-aarch64.so.1 (0x0000ffffb0110000)
```

.github/workflows/libqrmi.yml
```diff
@@ -321,6 +338,11 @@ jobs:
...
```
ohtanim: AFAIK, openssl-devel is no longer required. We elimimated the dependencies of openssl in #67.

Actually, ldd does not show any dependency on OpenSSL.

```
# ldd libqrmi.so 
linux-vdso.so.1 (0x0000ffffb014c000)
libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x0000ffffb00df000)
libm.so.6 => /lib64/libm.so.6 (0x0000ffffb003e000)
libc.so.6 => /lib64/libc.so.6 (0x0000ffffaee52000)
/lib/ld-linux-aarch64.so.1 (0x0000ffffb0110000)
```

.github/workflows/libqrmi.yml
```diff
@@ -387,4 +409,9 @@ jobs:
         name: ${{ inputs.artifact-prefix }}-libqrmi-${{ needs.info.outputs.qrmi-version }}-tarball-arm64
         path: libqrmi-${{ needs.info.outputs.qrmi-version }}*.tar.gz
         if-no-files-found: error
-        retention-days: 1
\ No newline at end of file
+        retention-days: 1
+    - if: inputs.libqrmi-rpm == true
+      name: Build libqrmi rpms
+      run: |
+        dnf install -y rpm-build gcc make openssl-devel pkg-config
```
ohtanim: AFAIK, openssl-devel is no longer required. We elimimated the dependencies of openssl in #67.

Actually, ldd does not show any dependency on OpenSSL.

```
# ldd libqrmi.so 
linux-vdso.so.1 (0x0000ffffb014c000)
libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x0000ffffb00df000)
libm.so.6 => /lib64/libm.so.6 (0x0000ffffb003e000)
libc.so.6 => /lib64/libc.so.6 (0x0000ffffaee52000)
/lib/ld-linux-aarch64.so.1 (0x0000ffffb0110000)
```

docker/manylinux/Dockerfile
```diff
@@ -13,6 +13,7 @@ ENV INSIDE_CONTAINER=1
 WORKDIR ${QRMI_DIR}
 
 RUN \
+  dnf install -y rpm-build gcc make openssl-devel pkg-config && \
```
ohtanim: AFAIK, openssl-devel is no longer required. We elimimated the dependencies of openssl in #67.

Actually, ldd does not show any dependency on OpenSSL.

```
# ldd libqrmi.so 
linux-vdso.so.1 (0x0000ffffb014c000)
libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x0000ffffb00df000)
libm.so.6 => /lib64/libm.so.6 (0x0000ffffb003e000)
libc.so.6 => /lib64/libc.so.6 (0x0000ffffaee52000)
/lib/ld-linux-aarch64.so.1 (0x0000ffffb0110000)
```

packaging/rpm/libqrmi.spec
```diff
@@ -0,0 +1,125 @@
...
+Source0:        qrmi-%{version}.tar.gz
+Source1:        qrmi-vendor-%{version}.tar.gz
+
+BuildRequires:  gcc
+BuildRequires:  make
+# Ensure the build host is RHEL 8 (glibc 2.28) or newer
+BuildRequires:  glibc >= 2.28
+# reqwest pulls in hyper-tls -> openssl-sys, which calls out to the
+# system openssl-devel headers and pkg-config to locate them at compile time.
+BuildRequires:  openssl-devel
```
ohtanim: AFAIK, openssl-devel is no longer required. We elimimated the dependencies of openssl in #67.

Actually, ldd does not show any dependency on OpenSSL.

```
# ldd libqrmi.so 
linux-vdso.so.1 (0x0000ffffb014c000)
libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x0000ffffb00df000)
libm.so.6 => /lib64/libm.so.6 (0x0000ffffb003e000)
libc.so.6 => /lib64/libc.so.6 (0x0000ffffaee52000)
/lib/ld-linux-aarch64.so.1 (0x0000ffffb0110000)
```

packaging/rpm/libqrmi.spec
```diff
@@ -0,0 +1,125 @@
...
+%global abi_major 0
+
+Name:           libqrmi
+Version:        %{version}
+Release:        1%{?dist}
+Summary:        Vendor-agnostic C shared library and header file for controlling quantum computing resources
+License:        Apache-2.0
+URL:            https://github.com/qiskit-community/qrmi
+Source0:        qrmi-%{version}.tar.gz
+Source1:        qrmi-vendor-%{version}.tar.gz
```
ohtanim: Isn't this qrmi-%{version}-vendor.tar.gz?

https://github.com/qiskit-community/qrmi/releases/download/0.24.4/qrmi-0.24.4-vendor.tar.gz

Makefile_common.mk
```diff
@@ -33,6 +33,8 @@ CMAKE_BUILD_TYPE = Release
 endif
 
 LIBQRMI_SO_PATH = target/$(CARGO_PROFILE_DIR)/libqrmi.so
+QRMI_SOURCE_TARBALL_PATH = $(DIST_DIR)/qrmi-$(QRMI_VERSION).tar.gz
+QRMI_VENDOR_TARBALL_PATH = $(DIST_DIR)/qrmi-vendor-$(QRMI_VERSION).tar.gz
```
ohtanim: Are you planning to change the filename that has been provided up to now?

So far, it has been distributed in the Git repository under the filename `qrmi-$(QRMI_VERSION)-vendor.tar.gz`. I consider changing it to be a form of external behavior change. I think it would be better to keep using the existing filename. This is exactly the point I raised in an earlier review comment.

packaging/rpm/libqrmi.spec
```diff
@@ -0,0 +1,125 @@
...
+# Build
+# ---------------------------------------------------------------------------
+%build
+# Source the Rust environment if installed via rustup
+[ -f "$HOME/.cargo/env" ] && . "$HOME/.cargo/env"
+
+# Rust's toolchain does not natively manage SONAME fields out-of-the-box for cdylibs.
+# Embed SONAME in the ELF so RPM's find-provides automatically emits
+# libqrmi.so.{abi_major}()(64bit) and find-requires on consumers matches it correctly.
+RUSTFLAGS="-Clink-arg=-Wl,-soname,libqrmi.so.%{abi_major}" make build
```
ohtanim, line 80: Are the Makefile and Makefile_common.mk from this Git repository really required?

(Editted)
My understanding is that the spec file is meant to be portable. As such, I had expected RPM packages to be built from the following three release artifacts that we publish in the Git repository.

https://github.com/qiskit-community/qrmi/releases/download/0.24.4/libqrmi-0.24.4-el8-x86_64.tar.gz
https://github.com/qiskit-community/qrmi/releases/download/0.24.4/qrmi-0.24.4-vendor.tar.gz
https://github.com/qiskit-community/qrmi/archive/refs/tags/0.24.4.tar.gz

If those files are required during the RPM build process, could you clarify why?

ohtanim (changes requested):

packaging/rpm/libqrmi.spec
cclaudio: > Are the Makefile and Makefile_common.mk from this Git repository really required?
> 

We could build the libqrmi using a cargo command line if you prefer. I used something like `make build` because future changes in the build procedure would probably require us to update only the Makefile. The CI workflows, libqrmi spec and developers would automatically pick the changes through the Makefile.

> (Editted) My understanding is that the spec file is meant to be portable. As such, I had expected RPM packages to be built from the following three release artifacts that we publish in the Git repository.
> 
> https://github.com/qiskit-community/qrmi/releases/download/0.24.4/libqrmi-0.24.4-el8-x86_64.tar.gz https://github.com/qiskit-community/qrmi/releases/download/0.24.4/qrmi-0.24.4-vendor.tar.gz https://github.com/qiskit-community/qrmi/archive/refs/tags/0.24.4.tar.gz
> 
> If those files are required during the RPM build process, could you clarify why?

There shoudl be a way to hardcode URLs in the SOURCE field of the spec file, if that's what we want to do. Typically, distros don't do that because they want to have hermetic builds. On the other hand, I believe local consumers (e.g. STFC) would prefer to have the flexibility to build from a custom tree or from released tarballs. `make libqrmi-rpm` provides that flexibility, it will create the source and vendor tarballs only if they don't exist. For example, if local consumers want to build the rpms from released tarballs, they could just download the source and vendor tarballs from github, otherwise, they could let `make libqrmi-rpm` generate the tarballs from the local tree.

As for pre-built artifacts, we can adopt the policy that makes more sense for upstream. However, ultimately, distros and local consumers have their own policy for including pre-built artifacts. For example, RHEL and Fedora usually prefer not to include pre-built artifacts because of provenance. Local consumers might be a little more flexible with that.

Makefile_common.mk
cclaudio: Good catch. Thanks.
Fixed.

packaging/rpm/libqrmi.spec
cclaudio: You're right. Fixed

.github/workflows/libqrmi.yml
cclaudio: Fixed.

docker/manylinux/Dockerfile
cclaudio: Fixed

packaging/rpm/libqrmi.spec
cclaudio: Fixed

.github/workflows/libqrmi.yml
cclaudio: Fixed.

.github/workflows/libqrmi.yml
cclaudio: Fixed.

packaging/rpm/libqrmi.spec
ohtanim: @cclaudio In this case, you need to describe the prerequisites to run with this spec file. At very least, 'docker' or 'podman' should be available to create a RPM.

I failed to generate a rpm on my minimum linux env.

.github/workflows/libqrmi.yml
ohtanim: not yet committed.

.github/workflows/libqrmi.yml
ohtanim: not yet committed

Makefile_common.mk
ohtanim: not yet committed

packaging/rpm/libqrmi.spec
ohtanim:
```
# rpmbuild -ba --define "_topdir $PWD/rpmbuild" --define "version 0.24.4" ./rpmbuild/SPECS/libqrmi.spec 
warning: Macro expanded in comment on line 6: %{version}, libqrmi.so.0)

setting SOURCE_DATE_EPOCH=1788912000
Executing(%prep): /bin/sh -e /var/tmp/rpm-tmp.DrE698
+ umask 022
+ cd /shared/sandbox/rpm_build/rpmbuild/BUILD
+ cd /shared/sandbox/rpm_build/rpmbuild/BUILD
+ rm -rf qrmi-0.24.4
+ /usr/bin/gzip -dc /shared/sandbox/rpm_build/rpmbuild/SOURCES/qrmi-0.24.4.tar.gz
+ /usr/bin/tar -xof -
+ STATUS=0
+ '[' 0 -ne 0 ']'
+ cd qrmi-0.24.4
+ /usr/bin/chmod -Rf a+rX,u+w,g-w,o-w .
+ tar xf /shared/sandbox/rpm_build/rpmbuild/SOURCES/qrmi-0.24.4-vendor.tar.gz
+ mkdir -p .cargo
+ cat
+ RPM_EC=0
++ jobs -p
+ exit 0
Executing(%build): /bin/sh -e /var/tmp/rpm-tmp.n8W8WZ
+ umask 022
+ cd /shared/sandbox/rpm_build/rpmbuild/BUILD
+ cd qrmi-0.24.4
+ '[' -f /root/.cargo/env ']'
+ . /root/.cargo/env
++ case ":${PATH}:" in
+ RUSTFLAGS=-Clink-arg=-Wl,-soname,libqrmi.so.0
+ make build
Makefile_common.mk:45: *** "No container engine found. Please install docker or podman.".  Stop.
error: Bad exit status from /var/tmp/rpm-tmp.n8W8WZ (%build)


RPM build errors:
    Macro expanded in comment on line 6: %{version}, libqrmi.so.0)

    Bad exit status from /var/tmp/rpm-tmp.n8W8WZ (%build)
```

cclaudio: I added a section `Packaging` section in INSTALL.md with instructions

cclaudio:
Feedback applied.
- Removed openssl-devel requirement
- Added a Packaging section in the INSTALL.md with instructions
- Test libqrmi packaging only for x86_64 in the weekly CI. At the moment, nobody has shown interest on packaging for arm64 and ppc64le.

INSTALL.md
```diff
@@ -38,6 +38,7 @@
     - [🦀 Rust API docs](#how-to-generate-rust-api-document)
     - [🐍 Pythoni API docs](#how-to-generate-python-api-document)
     - [©️ C API docs](#how-to-generate-c-api-document)
+  - [ Packaging](#packaging)
```
ohtanim: Please remove unneccesary space.

packaging/rpm/libqrmi.spec
ohtanim: Don't we need to add 
`Requires: glibc >= 2.28`
?

ohtanim (changes requested):

cclaudio:
Thank you again for your feedback @ohtanim 

I added two new commits to address your comments.

ohtanim (approved):
LGTM

PR #239, opened 2026-09-05
Add default implementation of QuantumResource

### LINKED ISSUE #236: Add Default Implementations to `QuantumResource`, and an `UnsupportedFunction` Error Variant

The current definition of the `QuantumResource` trait provides no default implementations for its methods. As a result, every vendor implementation must implement all trait functions, even for functions that a given vendor's
backend does not actually support.

For example, `acquire()` and `release()` are not meaningful operations for every vendor's backend, yet each implementation is currently required to provide some implementation for them regardless.

This gap is also noted as an area for improvement in the Oak Ridge National Laboratory (ORNL) analysis report comparing QRMI and QDMI:
https://github.com/openQSE/development-analysis/blob/main/qrmi-qdmi/qrmi-qdmi-analysis.md

Add a default implementation to every method on the `QuantumResource` trait. Each default implementation returns `Err(QrmiError::UnsupportedFunction(..))`. With this change, a vendor only needs to override (implement) the functions its backend actually supports; unsupported functions can simply be left at their default.

### New error variant

Add a new variant to `QrmiError` (in `src/error.rs`), following the existing pattern used by other variants in that file:

```rust
/// The requested operation is not supported by this backend's
/// implementation of `QuantumResource`.
#[error("unsupported function: {0}")]
UnsupportedFunction(String),
```

The variant carries the name of the unsupported function (e.g. `"acquire"`), so that callers (such as Slurm, via the C bindings) can identify which operation was not supported, rather than only seeing a generic error.

As with other variants, `UnsupportedFunction` is added to `QrmiError::kind()` and to `QrmiErrorKind`, so it is also exposed through the existing C/Python binding error-mapping mechanism.

### Scope

This applies to the methods defined on the `QuantumResource` trait itself. Constructor `new()` is out of scope here.

### Backward compatibility

Existing vendor implementations currently implement every trait method explicitly, whether or not the underlying backend supports the operation. Adding default implementations does not change this: existing implementations continue to compile unchanged, and are free to remove their own explicit (unsupported) implementations in favor of the new default at their own pace.

None.

ASAP(September 10 at latest). This is needed for a new QRMI vendor backend currently in preparation, which does not support all `QuantumResource` functions.

ohtanim:
## Description of Change

This PR contains an implementation of #236. 

- Added default implementation to the QuantumResource trait functions except `resource_id()`, `resource_type()` and `is_accessible()`.
- Default implementation of `acquire()` will return dummy acquisition ID with showing warning message for backward compatibility. This behavior will be changed to `Err(UnsupportedFunction)` in future release.
- Default implementation of `release()` is a noop function with showing warning message for backward compatibility. This behavior will be changed to return `Err(UnsupportedFunction)` in future release.
- Default implementation of metadata return an empty hashtable with showing warning message since we cannot return Error. This `metadata()` function will be replaced with other new function, so this proposed implementation is enough.
- Remove `acquire()`/`release()` functions from IBM Quantum System/IQM Server QRMI/Pasqal Cloud implementations. 
- Remove `task_log` from Alice and Bob Felis QRMI.

## Checklist ✅

## Ticket
- [x] Fixes #236 
- [ ] Is Part of #

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,8 +25,6 @@ pub use resource_provider::ResourceProvider;
 pub mod service;
 pub use service::QRMIService;
 
-use uuid::Uuid;
-
 mod cext;
 pub mod models;
 #[cfg(feature = "pyo3")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,6 +25,8 @@ pub use resource_provider::ResourceProvider;
 pub mod service;
 pub use service::QRMIService;
 
+use uuid::Uuid;
+
 mod cext;
 pub mod models;
 #[cfg(feature = "pyo3")]
@@ -91,6 +93,7 @@ use async_trait::async_trait;
 pub type Result<T> = std::result::Result<T, QrmiError>;
 
 /// Defines interfaces to quantum resources.
+#[allow(unused_variables)]
 #[async_trait]
 pub trait QuantumResource: Send + Sync {
     /// Returns resource identifier of this quantum resource.
@@ -108,7 +111,11 @@ pub trait QuantumResource: Send + Sync {
     ///     Ok(())
     /// }
     /// ```
-    async fn resource_id(&mut self) -> Result<String>;
+    async fn resource_id(&mut self) -> Result<String> {
+        Err(QrmiError::UnsupportedFunction(
+            "qrmi::QuantumResource::resource_id".to_string(),
+        ))
+    }
 
     /// Returns resource type of this quantum resource.
     ///
@@ -125,7 +132,11 @@ pub trait QuantumResource: Send + Sync {
     ///     Ok(())
     /// }
     /// ```
-    async fn resource_type(&mut self) -> Result<ResourceType>;
+    async fn resource_type(&mut self) -> Result<ResourceType> {
+        Err(QrmiError::UnsupportedFunction(
+            "qrmi::QuantumResource::resource_type".to_string(),
+        ))
+    }
 
     /// Returns true if device is accessible, otherwise false. A target quantum resource is not considered accessible if quantum workloads cannot be executed, even when the system itself is reachable, for example due to maintenance.
     ///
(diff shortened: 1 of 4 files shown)


MatthieuMoreau0 (approved):
lgtm, thank you!

src/pyext.rs
```diff
@@ -102,6 +108,7 @@ fn to_py_err(err: QrmiError) -> PyErr {
         | QrmiErrorKind::InvalidConfig => ConfigError::new_err(msg),
         QrmiErrorKind::UnsupportedResourceType => UnsupportedResourceTypeError::new_err(msg),
         QrmiErrorKind::UnsupportedPayload => UnsupportedPayloadError::new_err(msg),
+        QrmiErrorKind::UnsupportedFunction => UnsupportedFunctionError::new_err(msg),
```
awennersteen, line 111: Should this new one also be added to the bottom of this file like here:
https://github.com/qiskit-community/qrmi/blob/main/src/pyext.rs#L821
?

awennersteen (approved):
lgtm

src/pyext.rs
ohtanim: @awennersteen Good catch. Thank you. I forgot to replicate this change when I created this PR. Double checked.

awennersteen: Thanks

ohtanim:
@machjn Could you please review and approve? I also modified Alice & Bob QuantumResource code.

src/lib.rs
```diff
@@ -209,7 +228,11 @@ pub trait QuantumResource: Send + Sync {
...
     /// }
     /// ```
-    async fn task_start(&mut self, payload: Payload) -> Result<String>;
+    async fn task_start(&mut self, payload: Payload) -> Result<String> {
+        Err(QrmiError::UnsupportedFunction(
+            "qrmi::QuantumResource::task_start".to_string(),
+        ))
+    }
 
     /// Stops the task specified by `task_id`. This function is called if the user cancels the job or if the time limit for job execution is exceeded. The implementation must cancel the task if it is still running.
```
OkuyanBoga, line 237: Small documentation suggestion: The current “must cancel” wording could read as a requirement for every backend

OkuyanBoga (approved):
Looks good to me, thanks

src/lib.rs
ohtanim: @OkuyanBoga I believe this is correct. This function was defined to ask every backend implementation to cancel job execution if exists. Please let me know if there's any backend that doesn't fall under this. I don't think there is one. 

There have been comments about this code since the early days. You must be aware of that too.

ohtanim:
@machjn Since this was needed urgently, I'm going ahead and merging it.
Regarding the Felis QRMI changes, I removed the previously unimplemented `acquire`, `release`, and `task_logs` functionality and switched to the default implementation instead. Please keep that in mind.

PR #243, opened 2026-09-09
feat/`from_config()` interface

### LINKED ISSUE #234: Per-Instance Configuration for QRMI
## Background

The current QRMI implementation is configured via environment variables. This works well in typical distributed-processing environments, where the calling process forks a new process for each workload execution. It does not work, however, when a single process needs to use different QRMI configurations concurrently without relying on process-global state.

Upcoming Slurm feature(HRES Lua Pugin), Slurm existing features (Burst Buffer Lua plugin, and the Job Submit plugin) — invoke plugin code in one of two ways:

1. `slurmctld` calls some functions in the plugin interface directly, in-process.
2. A forked child process calls some other functions of the Lua plugin.

In case (2), the current environment-variable-based approach is not a problem, since each forked process has its own independent environment.

In case (1), however, calls happen sequentially within a single, long-lived `slurmctld` process (the plugin invocation itself is single-threaded, so there is no concern about concurrent access). The problem is that each call may require a *different* QRMI configuration. With the current design, the caller would need to set the relevant environment variables before every call and (ideally) clean them up afterward. This is undesirable because:

- `slurmctld` is central daemon of Slurm cluster. 
- It places the burden of managing process-global, mutable state on the caller (`slurmctld`), which must remember to set and unset environment variables around every call.
- If a call fails partway through, cleanup of the environment variables can easily be missed, leaving stale configuration for the next call. 
- It does not scale gracefully if Slurm's calling pattern changes in the future.

Establish from_config() as a standard constructor that each backend implementation provides, alongside the existing new():

```rust
pub fn from_config(config: HashMap<String, String>) -> Result<Self, QrmiError>;
```

Instead of reading configuration from environment variables, this constructor builds a QRMI instance directly from a `HashMap<String, String>` of key/value pairs supplied by the caller. This moves configuration from being **per-process** to being **per-instance**, which is exactly what is needed to support the `slurmctld`-direct-call pattern described above.

A few design notes:

- **Symmetry with `new()`, not a builder chain.** `from_config()` is a standalone constructor, symmetric with the existing `new()`, not a method meant to be chained onto it (i.e., not `new().with_config(...)`). The current `new()` performs eager validation and returns `Err` immediately if required environment variables are missing or invalid. A builder-style `new().with_config()` chain would risk making that validation lazy, since it's unclear at which step configuration should be checked. Keeping `from_config()` as an independent constructor preserves the existing fail-fast contract: both `new()` and `from_config()` validate their input
  and return a `Result` at the point of construction.
- **No hidden global state.** Aside from `env_logger`, the current implementation holds no global state related to QRMI configuration. This means instances constructed via `from_config()` can safely coexist with different configurations within the same process, with no risk of one instance's configuration leaking into another's.
- **Existing behavior is preserved.** The environment-variable-based `new()` constructor is left unchanged. It remains the right approach for traditional distributed-processing environments where each workload runs in its own forked process. This proposal only adds a new construction path; it does not change the existing trait signature or behavior.

**Setting/unsetting environment variables around each call.** Since Slurm's plugin invocation in the `slurmctld`-direct-call case is single-threaded, thread-local storage is not a relevant concern here. The realistic alternative is for the caller to set the required environment variables immediately before each call and unset them afterward. This was rejected because:

1. It places stateful setup/cleanup responsibility on the caller for every single call.
2. If an error occurs mid-call, cleanup can be skipped, leaving stale environment state for the next invocation.
3. It does not accommodate future changes to Slurm's calling pattern (e.g., if plugin invocation were to become concurrent).

`from_config()` avoids all of this by encapsulating configuration entirely within the instance itself.

## Scope

- This proposal targets all backends currently supported by QRMI. That said, initial implementation may reasonably start with the backend(s) belonging to the vendor(s) most interested in this Slurm feature, with the remaining
backends following in subsequent work.
- C API should be available at the same time.

## Expected availability

Availability by end of September is expected. This timeline allows for integration work with the upcoming Slurm release to begin promptly, so that functionality can be verified before that Slurm version is released.

badtst:
## Description of Change

Adds a `::from_config(HashMap<String, String>)` constructor alongside the existing `::new()` constructor for every QuantumResource implementation. This lets callers configure and instantiate a resource per-instance from an in-memory map instead of process-wide environment variables.

### Changed

- New `from_config constructor` per provider, each documenting its required/optional keys 
- src/common/create_resource.rs: new create_resource_from_config dispatches to the right provider's from_config by ResourceType, mirroring create_resource.
- C bindings (src/cext.rs): new `qrmi_resource_new_from_config`, taking a `QrmiConfigMap` which is an alias for the `QrmiEnvironmentVariables` struct.
- Python bindings (src/pyext.rs): new `PyQuantumResource.from_config` static method.
- Lua bindings (lua/lua_qrmi.c): new `qrmi.new_from_config` function to create a resource from a lua table. 
- src/error.rs: QrmiError::MissingConfigKey's message/docs were generalized from "environment map" to "config map", since the error is now shared between provider env-var maps and the new generic from_config maps.
- src/lib.rs: removed default (no-op/warn-and-continue) implementations from the QuantumResource trait's acquire/release/task_*/target/metadata methods, making them required — every implementor now provides its own behavior explicitly.

### Notes

- Regarding the config's keys, they are defined the same way as the environment variables but without the eventual `{backend_name}_` prefix. Checkout the migration documentation for more details.
- Checked that the new methods do appear in the auto-generated docs

### Examples

1. **Rust** — Call `T::from_config(resource_id, config)` instead of
   `T::new(resource_id)`:

   ```rust
   use qrmi::ibm::IBMQuantumComputeService;
   use std::collections::HashMap;

   let config = HashMap::from([
       ("QRMI_IBM_QCS_ENDPOINT".to_string(), "...".to_string()),
       ("QRMI_IBM_QCS_IAM_ENDPOINT".to_string(), "...".to_string()),
       ("QRMI_IBM_QCS_IAM_APIKEY".to_string(), "...".to_string()),
       ("QRMI_IBM_QCS_SERVICE_CRN".to_string(), "...".to_string()),
   ]);
   let qrmi = IBMQuantumComputeService::from_config("ibm_kingston", config)?;
   ```

2. **C** — Call the new `qrmi_resource_new_from_config()` instead of
   `qrmi_resource_new()`, building a `QrmiConfigMap` the same way you'd
   build any other `QrmiEnvironmentVariables`-shaped value:

   ```c
   QrmiKeyValue variables[] = {
       {(char *)"QRMI_WARDEN_URL", (char *)"http://localhost:8006"},
       {(char *)"QRMI_JOB_ID",     (char *)"1"},
       {(char *)"QRMI_JOB_UID",    (char *)"1000"},
   };
   QrmiConfigMap config = { .variables = variables, .length = 3 };

   QrmiQuantumResource *qrmi = qrmi_resource_new_from_config(
       "PASQAL_LOCAL", QRMI_RESOURCE_TYPE_PASQAL_LOCAL, &config);
   ```

   As with `qrmi_resource_new()`, the returned handle must be released with
   `qrmi_resource_free()` once it's no longer needed.

3. **Python** — Call the new `QuantumResource.from_config()` static method
   instead of the `QuantumResource(resource_id, resource_type)` constructor:

   ```python
   from qrmi import QuantumResource, ResourceType

   qrmi = QuantumResource.from_config(
       "ibm_kingston",
       ResourceType.IBMQuantumComputeService,
       {
           "QRMI_IBM_QCS_ENDPOINT": "...",
           "QRMI_IBM_QCS_IAM_ENDPOINT": "...",
           "QRMI_IBM_QCS_IAM_APIKEY": "...",
           "QRMI_IBM_QCS_SERVICE_CRN": "...",
       },
   )
   ```

4. **Lua** — Call the new `qrmi.new_from_config()` instead of `qrmi.new()`,
   passing a plain Lua table (string keys/values).

   ```lua
   local resource, err = qrmi.new_from_config("ibm_kingston", "ibm-quantum-compute-service", {
       QRMI_IBM_QCS_ENDPOINT = "...",
       QRMI_IBM_QCS_IAM_ENDPOINT = "...",
       QRMI_IBM_QCS_IAM_APIKEY = "...",
       QRMI_IBM_QCS_SERVICE_CRN = "...",
   })
   ```

### TODO

- [x] Find standard for naming config parameters from env variable names
- [x] Introduce new C types for config
- [x] Adapt error messages to not only mention env variables source
- [x] Lua bindings

## Checklist ✅

## Ticket
- [x] Fixes #234 
- [ ] Is Part of #


badtst:
@ohtanim I think I'm almost done for this, just a few points where I'd like your input:
- For the moment I tried to match the config parameters names to env variable suffixes so that they could be parsed in the future
- `backend_name`/`resource_id` should be passed within the `config` object or as a separated parameter for `from_config` similarly to `new` ? Or within the `config` hashmap itself ? If so we'd need to make a choice between `backend_name` and `resource_id` to homogenize accross resources/providers
- Should I move ahead with lua bindings or do you wish to implement them in a separate PR .
- Once we are done, how should we document this new interface ? Do we duplicate the examples ? Or just mention it in the `.md` files ?

Lmk what you think so far

ohtanim:
@badtst Thank you for prompt actions. Great!

> * `backend_name`/`resource_id` should be passed within the `config` object or as a separated parameter for `from_config` similarly to `new` ? Or within the `config` hashmap itself ? If so we'd need to make a choice between `backend_name` and `resource_id` to homogenize accross resources/providers

Let's keep it consistent with new() and pass it as a separate parameter.

BTW, for Python binding, user can specify HashMap<String, String> as argument, correct ?
For C, I assume user can pass QrmiEnvironmentVariables like below. Correct?

```
  QrmiEnvironmentVariables envvars;

  envvars.variables = malloc(sizeof(QrmiKeyValue) * 2);
  envvars.length = 2;
  envvars.variables[0].key = "k0";
  envvars.variables[0].value = "v0";
  envvars.variables[1].key = "k1";
  envvars.variables[1].value = "v1";
```

> * Should I move ahead with lua bindings or do you wish to implement them in a separate PR .

If you'd like to take it on, that would be great. Any interest in the Lua bindings?

> * Once we are done, how should we document this new interface ? Do we duplicate the examples ? Or just mention it in the `.md` files ?

Describing in the `.md` is enough to me. As this is likely to be part of the 0.25.0 release, could you please create docs/migration/0.25.0.md and document this change there? Can you confirm that the new functions will appear in Rustdoc, Doxygen, and pydoc? I would expect that to work as long as the source code is documented in the same way as the existing functions.

badtst:
> Let's keep it consistent with new() and pass it as a separate parameter.

Thanks I also think it's best that way, I'll adapt the methods.

> BTW, for Python binding, user can specify HashMap<String, String> as argument, correct ?

Here is an example of what it would look like in python (we can directly use a python `dict`)

```python
config = {"backend_name": "FRESNEL_CAN1", "config_root": "/workspace", "project_id": PROJECT_ID}
qrmi = QuantumResource.from_config(ResourceType.PasqalCloud, config)
```

For C I created a type alias for `QrmiEnvironmentVariables` called `QrmiConfigMap` as I feel the naming aligns better with the usage, but it works the same. Here is what it would look like:

```c
    QrmiConfigMap config;
    
    QrmiKeyValue variables[] = {
        {(char *)"backend_name", (char *)"PASQAL_LOCAL"},
        {(char *)"warden_url", (char *)"http://localhost:8006"},
        {(char *)"job_id", (char *)"1"},
        {(char *)"job_uid", (char *)"1000"},
    };
    config.variables = variables;
    config.length = 4;
    
    QrmiQuantumResource *qrmi = qrmi_resource_new_from_config(QRMI_RESOURCE_TYPE_PASQAL_LOCAL, &config);

```

> If you'd like to take it on, that would be great. Any interest in the Lua bindings?

I have no particular experience/interest in lua for the moment. If I have the time to dedicate to this I'll do it probably since it is probably goinf to be needed for future plugins.

> Describing in the .md is enough to me. As this is likely to be part of the 0.25.0 release, could you please create docs/migration/0.25.0.md and document this change there? Can you confirm that the new functions will appear in Rustdoc, Doxygen, and pydoc? I would expect that to work as long as the source code is documented in the same way as the existing functions.

I'll check all of this thx !

ohtanim:
@badtst Tested with C example
- IQM .. Succeeded
- IBM Quantum Compute Service .. Succeeded
- IBM Quantum System .. Failed
   - example with environment variables .. Succeeded
   - example with new_from_config .. Failed

1) I made the following changes to examples/qrmi/c/ibm_quantum_system/src/quantum_system.c

```
#if 0
  load_dotenv();

  QrmiQuantumResource *qrmi =
      qrmi_resource_new(argv[1], QRMI_RESOURCE_TYPE_IBM_QUANTUM_SYSTEM);
#else
QrmiKeyValue variables[] = {
    {(char *)"endpoint", (char *)"http://192.168.1.9:8080"},
    {(char *)"iam_endpoint", (char *)"https://iam.test.cloud.ibm.com"},
    {(char *)"iam_apikey", (char *)"my iam key"},
    {(char *)"service_crn", (char *)"my instance"},
    {(char *)"aws_access_key_id", (char *)"my access key id"},
    {(char *)"aws_secret_access_key", (char *)"my secret access key"},
    {(char *)"s3_endpoint", (char *)"https://s3.us-east.cloud-object-storage.appdomain.cloud"},
    {(char *)"s3_bucket", (char *)"cloud-object-storage-cos-standard-5sn"},
    {(char *)"s3_region", (char *)"us-east"},
};
QrmiConfigMap config = { .variables = variables, .length = 4 };

QrmiQuantumResource *qrmi = qrmi_resource_new_from_config(
    argv[1], QRMI_RESOURCE_TYPE_IBM_QUANTUM_SYSTEM, &config);
#endif
```

2) and run against real hardware.
```
$ ./quantum_system test_heron /shared/job_scripts/input/sampler_input_test_heron_params_only.json sampler
[2026-09-15T12:23:58Z ERROR qrmi::cext] S3 bucket is not configured. Use ClientBuilder.with_s3_bucket() to use this function.
failed to start a task. S3 bucket is not configured. Use ClientBuilder.with_s3_bucket() to use this function. (100)
```

Am I missing anything important?

badtst:
Hello Ohtani-san. I did test locally with `pasqal-local` and `pasqal-cloud`. My guess in your case is that you need to adjust the `length` parameter of the `QrmiConfigMap` here:
> QrmiConfigMap config = { .variables = variables, .length = 4 };

Otherwise when building the hashmap from the configmap, some parameters will be ignored in `cext::envvars_to_hashmap()`

Maybe we can design a builder function to avoid this error from users in the future ?

ohtanim:
> Hello Ohtani-san. I did test locally with `pasqal-local` and `pasqal-cloud`. My guess in your case is that you need to adjust the `length` parameter of the `QrmiConfigMap` here:
> 
> > QrmiConfigMap config = { .variables = variables, .length = 4 };
> 
> Otherwise when building the hashmap from the configmap, some parameters will be ignored in `cext::envvars_to_hashmap()`
> 
> Maybe we can design a builder function to avoid this error from users in the future ?

Thanks. You're right. I should change the length to 9.

ohtanim:
Another issue is found during IBM Quantum System QRMI.

Even though I use new_from_config(), I still needed to set QRMI_JOB_TIMEOUT_SECONDS environment variable to invoke task_start().

```
[2026-09-15T13:39:07Z ERROR qrmi::cext] test_heron_QRMI_JOB_TIMEOUT_SECONDS environment variable is not set
failed to start a task. test_heron_QRMI_JOB_TIMEOUT_SECONDS environment variable is not set (102)
```

I will revisit this fix once this PR is merged to main.

badtst:
Nice catch indeed, I did not see it being required in the `task_start` itself

ohtanim:
Tested with Lua example

- IQM .. Succeeded
- IBM Quantum Compute Service .. Succeeded
- IBM Quantum System .. Succeeded

Tested with Python example

- IQM .. Succeeded
- IBM Quantum Compute Service .. Succeeded
- IBM Quantum System .. Succeeded

src/ibm/qiskit_runtime_service.rs
```diff
@@ -107,6 +107,64 @@ impl IBMQiskitRuntimeService {
...
+    /// * `iam_endpoint` - IAM endpoint URL
+    /// * `iam_apikey` - IAM API key for QRS
+    /// * `service_crn` - QRS service instance CRN
+    ///
+    /// # Optional keys
+    ///
+    /// * `session_mode` - Session mode (default: "dedicated")
+    /// * `session_max_ttl` - Session max_ttl (default: 28800)
+    /// * `timeout_seconds` - Cost for the job (seconds)
+    /// * `session_id` or `job_acquisition_token` - pre-set session ID
```
ohtanim: Oh.. job_acquisition_token.

For the keys specified in the current config map, my understanding is that they are generally expected to be the lowercase form of the environment variable names defined by each vendor, with the vendor-specific prefix removed.

However, in addition to vendor-defined variables, there are also common environment variables provided by components such as the SPANK plugin:
- QRMI_JOB_TIMEOUT_SECONDS
- QRMI_JOB_ACQUISITION_TOKEN

Suppose a vendor were to define variables such as:
- QRMI_IBM_QS_JOB_ACQUISITION_TOKEN
- QRMI_IBM_QS_JOB_TIMEOUT_SECONDS
How should these be distinguished from the common variables above? It seems that this would break the current naming convention for keys in the config hashmap, since both the vendor-specific and common variables could map to similar key names.

badtst: That was one of my worries indeed.

The more I think about it, the more I think the best would be to:
- Keep the common parameters "as-is" meaning: `QRMI_JOB_TIMEOUT_SECONDS` env variable -> `qrmi_job_timeout_seconds` key
- Just remove the resource_id prefix from vendor variables : `<RESOURCE_ID>_QRMI_IBM_QS_ENDPOINT` env variable -> `qrmi_ibm_qs_enddpoint` key

This might be verbose but it would closely match the "environment" key structure from `qrmi_config.json` and make the parsing from environment variables/`qrmi_config.json` to `from_config()` hashmaps key/value pairs easier. 

What do you think ?

ohtanim: It might be a bit inelegant, but how about using the environment variable names as-is without converting them to lowercase? For example, QRMI_IBM_QS_ENDPOINT.

If we provide a wrapper function around env::var(), I don't think we'd need all the additional conversion logic that's currently required.

For example:
```
fn resolve(config: &Option<HashMap<String, String>>, key: &str) -> Result<String, env::VarError> {
    match config {
        Some(map) => map.get(key).cloned().ok_or(env::VarError::NotPresent),
        None => env::var(key),
    }
}
```
With a wrapper like this, we could simply replace calls to `std::env::var()` mechanically, for example:
`foo: resolve(&config, "FOO").unwrap_or_default(),`
and that should be sufficient.

ohtanim: Of course, the {resource_id}_ prefix would be skipped in the code above, and the resulting environment variable name would then be used.
For example, if the resource ID is foo, then:
ibm_kingston_QRMI_IBM_QS_ENDPOINT
would be mapped to:
QRMI_IBM_QS_ENDPOINT
before being passed to resolve().
This way, the code can continue to use a single canonical key name regardless of the resource ID, and the migration from std::env::var() to resolve() remains a straightforward mechanical replacement

badtst: Having this resolve wrapper could be useful indeed. 

We could also lowercase the key input when checking for the config keys 

I'll check this out tomorrow

ohtanim:
@badtst Can I resume review? If so, please re-request review.

lua/lua_qrmi.c
```diff
@@ -174,6 +174,138 @@ static int l_qrmi_new(lua_State *L) {
...
+        pairs[i].value = strdup(lua_tostring(L, -2));
+        i++;
+        lua_pop(L, 2); /* pop key copy and value, leaving key for lua_next */
+    }
+
+    return pairs;
+}
+
+/**
+ * @brief `qrmi.new_from_config(resource_type, config)` - Create a quantum
```
ohtanim: Is resource_id missing from the function arguments?

lua/lua_qrmi.c
```diff
@@ -174,6 +174,138 @@ static int l_qrmi_new(lua_State *L) {
...
+    }
+
+    size_t i = 0;
+    lua_pushnil(L);
+    while (lua_next(L, idx) != 0) {
+        /* Copy the key before converting it to a string: lua_tostring may
+         * rewrite the value in place on the stack, which would confuse the
+         * next lua_next() call if done directly on the traversal key. */
+        lua_pushvalue(L, -2);
+        if (!lua_isstring(L, -1) || !lua_isstring(L, -2)) {
```
ohtanim: According to [this document](https://www.lua.org/manual/3.1/manual.html), lua_isstring accepts strings and numbers. Doesn't your code assume string value only?

If you assume string value only, then
`lua_type(L, -1) == LUA_TSTRING`
must be correct way.

lua/lua_qrmi.c
```diff
@@ -174,6 +174,138 @@ static int l_qrmi_new(lua_State *L) {
...
+            free_config_pairs(pairs, i);
+            luaL_error(L, "config table keys and values must be strings");
+        }
+        pairs[i].key = strdup(lua_tostring(L, -1));
```
ohtanim, line 239: I would recommend NULL check for strdup().

lua/lua_qrmi.c
```diff
@@ -174,6 +174,138 @@ static int l_qrmi_new(lua_State *L) {
...
+        luaL_error(L, "out of memory building config map");
+    }
+
+    size_t i = 0;
+    lua_pushnil(L);
+    while (lua_next(L, idx) != 0) {
+        /* Copy the key before converting it to a string: lua_tostring may
+         * rewrite the value in place on the stack, which would confuse the
+         * next lua_next() call if done directly on the traversal key. */
+        lua_pushvalue(L, -2);
```
ohtanim, line 234: Very good!

lua/lua_qrmi.c
```diff
@@ -174,6 +174,138 @@ static int l_qrmi_new(lua_State *L) {
...
+ * `key`/`value` strings) and must free it with free_config_pairs().
+ *
+ * @param L Lua state.
+ * @param idx Stack index of the table to convert.
+ * @param[out] out_count Set to the number of entries in the returned array.
+ * @return Newly allocated array of @p *out_count QrmiKeyValue entries
+ *         (NULL if the table is empty). Never returns on failure (raises a
+ *         Lua error instead).
+ */
+static QrmiKeyValue *build_config_map_from_table(lua_State *L, int idx, size_t *out_count) {
```
ohtanim, line 208: do we accept negative value of idx? if negative value is specified, below code doesn't work as expected, I think.

src/alice_bob/felis.rs
```diff
@@ -37,23 +36,56 @@ impl AliceBobFelis {
...
+    }
+
+    /// Shared parsing and client-building logic for [`Self::new`]
+    /// (`config: None`, reads OS environment variables, optionally prefixed
+    /// by the backend name) and [`Self::from_config`] (`config: Some`,
+    /// reads the given map).
+    fn from_opt(backend_name: &str, config: Option<&HashMap<String, String>>) -> Result<Self> {
+        // Env vars are per-instance (`<backend_name>_QRMI_...`); config map
+        // keys use the same name minus that prefix.
+        let prefix = if config.is_some() {
```
ohtanim, line 67: `if config.is_some()` is done twice in this function.

src/alice_bob/felis.rs
```diff
@@ -37,23 +36,56 @@ impl AliceBobFelis {
...
+            QrmiError::MissingConfigKey
+        } else {
+            QrmiError::EnvVarNotSet
+        };
         // Handle environment variables
-        let api_key = env::var(format!("{backend_name}_QRMI_AB_FELIS_API_KEY"))
-            .or(env::var("QRMI_AB_FELIS_API_KEY"))
-            .map_err(|_| {
-                QrmiError::EnvVarNotSet(format!(
+        let api_key = resolve_opt(&format!("{prefix}QRMI_AB_FELIS_API_KEY"), config)
```
ohtanim: Having to write this much code for each parameter seems a bit excessive. Could we revisit the utility design and see if there's a way to reduce the boilerplate?
I'll also experiment with an implementation on my side later.

src/error.rs
```diff
@@ -214,11 +219,36 @@ pub enum QrmiErrorKind {
     Other,
 }
 
+/// Looks up `key` from `config` if given, otherwise from the OS environment.
+/// See [`required_env`]/[`resolve_opt_required`] for the mandatory case.
+pub(crate) fn resolve_opt(key: &str, config: Option<&HashMap<String, String>>) -> Option<String> {
+    match config {
+        Some(map) => map
+            .get(key)
+            .or_else(|| map.get(&key.to_lowercase()))
```
ohtanim: Why do we need to_lowercase() here?
I assume you're going to support Pascal Case or Camel Case?

ohtanim (changes requested):
I haven't yet covered all the changes, but I think it would be better to provide comments now.

lua/lua_qrmi.c
badtst: Yes missing from the description

src/error.rs
badtst: I documented it in the migration document (and most of the provider's also if I remember):

> Each resource type's `from_config` accepts the same keys as its `new()`'s
environment variables, minus only the eventual `{backend_name}_` prefix.
Every key also accepts its fully-lowercased form as a fallback
(e.g. `qrmi_ibm_qs_endpoint` works alongside `QRMI_IBM_QS_ENDPOINT`).

I figured it made more sense from an API perspective to accept snakecase when we are instantiating the QuantumResources from a script, and to accept the uppercase/env var name equivalent when we are going to be instantiating the QuantumResources from a configuration file such as `qrmi_config.json`

But I did not plan to support Pascal Case nor Camel case, even though it's doable in my opinion

src/alice_bob/felis.rs
badtst: `felis.rs` is the only provider with this kind of double env name parsing. The other providers better fit the utility design which we can still redesign

lua/lua_qrmi.c
badtst: True, numbers will be coerced into strings. From an API perspective we can force only strings yes

lua/lua_qrmi.c
badtst: True, idx should be a positive. Since it's an internal and only called with `idx = 3` I did not add a check but I'll add it for future reference

src/ibm/quantum_system.rs
badtst, line 173: There is duplicated environment parameters fetching happening in `task_result` and `task_logs`. We can address those in a separate PR imho

ohtanim: Agree.

I implemeted lazy configuration in IBM On-premise access because S3 parameters are only required for job execution. No need to specify in the context of scheduler integrations. Just fyi.

ohtanim: I will own those changes once this PR is completed.

badtst:
@ohtanim I addressed most of your comments, feel free to re-review whenever you have time

MatthieuMoreau0 (approved):
lgtm, thanks a lot!

ohtanim (approved):
LGTM. Thank you.

ohtanim:
@machjn This PR modified your Alice & Bob code. Could you please review and approve ? We will wait for your completion.

PR #245, opened 2026-09-11
Add status() method to QuantumResource trait for detailed resource status reporting

### LINKED ISSUE #244: Add status() method to QuantumResource trait for detailed resource status reporting

(Editted - Reconsidered the default value)
(Added - healthy)

## Summary

The QuantumResource trait currently exposes `is_accessible()`, which only reports resource availability as a binary value:

- True: Online and available
- False: Offline, or unavailable due to maintenance (e.g. calibration)

However, as noted in [recent paper](https://arxiv.org/abs/2607.19591), the actual state of a quantum resource has more nuance than this binary representation captures, and there is currently no way to query it. A human-readable status message would also be valuable in operational settings.

This PR proposes adding a `status()` method to the QuantumResource trait to expose more detailed state information, and deprecates the existing `is_accessible()`.

## Motivation

- A binary `is_accessible()` cannot distinguish between "offline and unavailable," "temporarily unavailable due to maintenance," and "online but effectively unusable due to contention from other jobs"
- Exposing more granular status lets operators and users make better decisions about job submission and troubleshooting 
- A free-form status message allows backends to communicate additional operational context (e.g. estimated time until maintenance completes)

Add `status()` to the `QuantumResource` trait:

```rust
#[async_trait]
pub trait QuantumResource {
    // ...existing methods...

    /// Returns detailed status information for the resource.
    /// Default implementation delegates to the (deprecated)
    /// `is_accessible()`, mapping `true` to `ONLINE`
    /// and `false` to `OFFLINE`, with `status_reason`
    /// set to `None`. Vendor implementations are expected to override
    /// this method to report `PAUSED` and `BUSY` as well, where
    /// supported.
    async fn status(&self) -> Result<ResourceStatus> {
        #[allow(deprecated)]
        let accessible = self.is_accessible().await?;
        Ok(ResourceStatusCode {
            status: if accessible {
                ResourceStatusCode::ONLINE
            } else {
                ResourceStatusCode::OFFLINE
            },
            status_reason: None,
            healthy: Some(true),
        })
    }
}
```

`ResourceStatus` has the following fields and methods:

```rust
pub struct ResourceStatus {
    pub status: ResourceStatusCode,
    /// Status reason such as "maintenance", "calibration". Vendor specific.
    pub status_reason: Option<String>,
    /// Whether the quantum computer is healthy.
    pub healthy: Option<bool>,　
}
```

`ResourceStatusCode` has the following values:

| Value | Description |
|---|---|
| `ONLINE` | Online and available |
| `OFFLINE` | Offline. Unavailable |
| `PAUSED` | Online but not currently accepting/running jobs due to a maintenance-type event (e.g. calibration), |
| `BUSY` | Online and available, but occupied by other users' jobs or has a large pending queue, so a submitted job will not run immediately |

## Vendor comparisons

| Vendors | status values | status reason is available? | healthy status is available? | busy |
| --- | --- | --- | --- | --- |
| IBM | `online`/`offline`/`paused` | yes | no, but might be able to generate from last_update_date in backend properties | locked flag (Quantum System API), dedicated session in use (Quantum Compute API)
| Pasqal | `ACTIVE`, `INACTIVE`, `RETIRED` | no | no | |
| Alice & Bob | yes. List Target Availabilities - enabled(bool) | may be yes. List Target Availabilities - message | yes, Check Health (bool value) | |
| AQT(ARNICA) | `maintenance`, `offline`, `online`, `unavailable` | no | | | 
| IQM | `online`/`maintenance` | no | yes (`healthy` - bool) | queue_length > N then BUSY? |
| OQTOPUS | `available`/`unavailable` | no | no, but might be able to generate from calibrated_at ? | n_pending_jobs > N then BUSY? | 
| IonQ | bool(available/unavailable) | no | yes (`degraded` - bool) | average_queue_time > T then BUSY? |
| Braket | `ONLINE`/`OFFLINE`/`RETIRED` | no  | no | queueSize > N then BUSY? |
| Quantinuum(Azure) | `Available`/`Degraded`(= paused)/`Unavailable` | no | unavailable | unknown |
| Rigetti | Unknown | | | |

## Vendor Implementation Requirement

- `status()` is a new trait method, but comes with a default implementation that delegates to the vendor's existing `is_accessible()`, so this is not a breaking change on its own
- By default, `status()` reports only `ONLINE` (when `is_accessible()` returns `true`) or `OFFLINE` (when it returns `false`), with `status_message` set to `None`
- Existing vendor implementations are expected to override `status()` after this PR merges in order to report `PAUSED` and `BUSY` where supported, and to provide a meaningful `status_reason`

## Deprecation of `is_accessible()`

- `QuantumResource::is_accessible()` is marked `#[deprecated]`
- Behavior is unchanged (returns `true` for ONLINE or BUSY)
- A default implementation is provided on `QuantumResource` itself, delegating to `status()`, to minimize changes required by each vendor implementation
- Removal of `is_accessible()` in a future major version will be considered

## Backward Compatibility

- `is_accessible()` will emit a deprecation warning but continue to function via each vendor's current implementation, unchanged
- `status()` is a new trait method with a default implementation that delegates to the deprecated `is_accessible()`, so existing vendor implementations continue to compile and run without modification until they choose to override it

## Open questions
- Do we prefer "Available", "Unavailable" and "Busy" instead of above 4 states ?

## Expected availability
Availability by end of September is expected. This timeline allows for integration work with the upcoming Slurm release to begin promptly, so that functionality can be verified before that Slurm version is released.

None.

- Update API document and inform users the deplication of is_accessible().

ohtanim:
This was discussed in #244 

### Summary of Added APIs — `ResourceStatus`

This summarizes every function and struct added or modified across the Rust core, Python bindings, C bindings, and Lua bindings as part of this PR.

#### Rust (core / `models::ResourceStatus`)

**`QuantumResource` trait**

- `async fn status(&mut self) -> Result<ResourceStatus>` — new. The default implementation delegates to the deprecated `is_accessible()`, returning only `Online/Offline`.
- `async fn is_accessible(&mut self) -> Result<bool>` — existing method, now marked `#[deprecated]` (behavior unchanged).

**`models/resource_status.rs` (new file)**

- `struct ResourceCapacity { available_slots: u32, max_slots: u32 }`
- `enum ResourceStatusCode { Online, Offline, Paused, Busy }`
- `struct ResourceStatus { status, status_reason: Option<String>, healthy: Option<bool>, capacity: Option<ResourceCapacity>, pending_job_count: Option<u64> }`
- `impl ResourceStatus { pub fn is_accessible(&self) -> bool }` — true when `ONLINE` or `BUSY`.
- `impl ResourceStatusCode { pub fn as_str(&self) -> &'static str }` —  `"online"` / `"offline"` / `"paused"` / `"busy"`.
- All three structs/enum derive `serde::Serialize`; `ResourceStatusCode` additionally has `#[serde(rename_all = "lowercase")]`.
- `ResourceStatus` is marked `/// cbindgen:ignore` (to avoid a name clash with `cext::ResourceStatus`).

#### Python (`pyo3`, `#[pymethods]` in `resource_status.rs` + `pyext.rs`)

**`ResourceStatus` (pyclass, `get_all`)**

- `is_accessible()` — `py_is_accessible`, exposed as `is_accessible` via `#[pyo3(name = "is_accessible")]`, thinly wraps the Rust `ResourceStatus::is_accessible()`.
- `to_dict() -> dict` — returns a JSON-compatible dict via `pythonize::pythonize()`.

**`PyQuantumResource` (`pyext.rs`)**

- `status()` — calls `self.qrmi.status()` and returns a `ResourceStatus`. - `is_accessible()` — existing method, now emits a `DeprecationWarning` (via `warnings.call_method1`).

**`m.add_class` registrations**

- `ResourceStatus`, `ResourceStatusCode`, `ResourceCapacity` newly registered.

#### C (`cext.rs`, opaque-handle pattern)

**New struct (opaque handle wrapping `models::ResourceStatus`)**

- `ResourceStatus { inner: crate::models::ResourceStatus }` → renamed to `QrmiResourceStatus` via cbindgen.

**New functions**

| Function | Signature | Notes |
|---|---|---|
| `qrmi_resource_status` | `(qrmi: *mut QuantumResource, outp: *mut *mut ResourceStatus) -> ReturnCode` | Obtains the handle |
| `qrmi_resource_status_free` | `(ptr: *mut ResourceStatus) -> ReturnCode` | Frees the handle |
| `qrmi_resource_status_code` | `(status, outp: *mut ResourceStatusCode) -> ReturnCode` | |
| `qrmi_resource_status_reason` | `(status) -> *mut c_char` | Nullable; free with `qrmi_string_free` |
| `qrmi_resource_status_healthy` | `(status, outp: *mut bool) -> ReturnCode` | `None` maps to `UNSUPPORTED_FUNCTION_ERROR` |
| `qrmi_resource_status_pending_job_count` | `(status, outp: *mut u64) -> ReturnCode` | Same as above |
| `qrmi_resource_status_capacity` | `(status, outp: *mut *mut ResourceCapacity) -> ReturnCode` | Same as above |
| `qrmi_resource_capacity_free` | `(ptr: *mut ResourceCapacity) -> ReturnCode` | Frees the handle |
| `qrmi_resource_status_is_accessible` | `(status, outp: *mut bool) -> ReturnCode` | |
| `qrmi_resource_status_code_to_string` | `(code: ResourceStatusCode) -> *const c_char` | Static string; no free needed |

**Changes to existing functions**

- `qrmi_resource_is_accessible` — Doxygen comment now includes `@deprecated` (behavior unchanged).

#### Lua (`lua_qrmi.c`)

- `static const char *status_code_to_string(QrmiResourceStatusCode s)` — internal helper.
- `static int l_status(lua_State *L)` — implements `resource:status()`; calls the C functions above and flattens the result into a plain Lua table (`healthy` / `pending_job_count` / `capacity` are converted from   `UNSUPPORTED_FUNCTION_ERROR` to `nil`).
- Registered `{"status", l_status}` in `resource_methods[]`.

#### Documentation

- Added rustdoc, doxygen, pydoc comments to the source. 
- `lua/API_REFERENCE.md` — added a deprecation note to `resource:is_accessible()` and a new `resource:status()` section.

####  Examples updated
- Updated `examples/qrmi/lua/*/example.lua` — replaced `resource:is_accessible()` with `resource:status()`.
- Updated `examples/qrmi/python/{alice_bob_felis, ibm_quantum_compute_service, ibm_quantum_system, iqm_server, pasqal_cloud, pasqal_local}/example.py - replaced `is_accessible()` with `status()`
- Added `examples/qrmi/python/resource_status
- Updated `examples/qrmi/rust/*` to use `qrmi.status().await?.is_accessible()`
- Added `examples/qrmi/c/resource_status`

## Checklist ✅

## Ticket
- [x] Fixes #244 
- [ ] Is Part of #


yoonho (commented):
The changes look ok to me BUT I do not believe I am qualified to give a full approval. I think one of the core maintainers should sign off on the changes.

examples/qrmi/python/resource_status/example.py
```diff
@@ -0,0 +1,87 @@
...
+        logger.info("offline")
+    case ResourceStatusCode.Paused:
+        logger.info("paused")
+    case ResourceStatusCode.Busy:
+        logger.info("busy")
+
+logger.info("reason: %s", status.status_reason)
+
+if status.pending_job_count is None:
+    logger.warning("this resource not report pending_job_count")
```
badtst:
```suggestion
    logger.warning("this resource does not report pending_job_count")
```

src/cext.rs
```diff
@@ -862,16 +872,451 @@ pub unsafe extern "C" fn qrmi_resource_is_accessible(
...
+}
+
+/// @ingroup QrmiQuantumResource
+/// Returns the status of the specified resource.
+///
+/// # Safety
+///
+/// * `qrmi` must have been returned by a previous call to qrmi_resource_new().
+///
+/// * `outp` must be non nul.
```
badtst: Repeated typo ?

```suggestion
/// * `outp` must be non-null.
```

src/models/resource_status.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Returns whether a job can currently be submitted to the resource.
+    /// Matches the previous behavior of `is_accessible()`
+    /// (true when Online or Busy).
+    pub fn is_accessible(&self) -> bool {
+        matches!(
+            self.status,
+            ResourceStatusCode::Online | ResourceStatusCode::Busy
+        )
+    }
+}
```
badtst, line 103: Is this is_accessible function going to be used in the future plugins or is it just there during the deprecation period of the `is_accessible` interface and for backward compatibility with the spank plugin ?

In the first case, should we consider a resource that is not `healthy` to be available ? Or maybe the `unhealthy` status should be moved to a new ResourceStatusCode ?

badtst (commented):
Overall every seems to be working on my side, just some questions on the impact of the `healthy` field

src/models/resource_status.rs
ohtanim: I don't expect future plugins to use this, but it might still be worth providing a utility equivalent to the previous is_accessible() behavior, that is, checking accessibility independently of whether the resource is healthy.

src/cext.rs
ohtanim: Fixed all.

src/models/resource_status.rs
ohtanim: > Or maybe the unhealthy status should be moved to a new ResourceStatusCode ?

I wonder if the root cause is that we put BUSY into ResourceStatusCode. As soon as we add UNHEALTHY, we end up having to define the semantics of combinations such as:
- BUSY AND UNHEALTHY
- ONLINE AND UNHEALTHY

It might be cleaner to reserve ResourceStatusCode for the resource state itself and expose busy-ness through a separate `busy` field, just as we do for `healthy`. That would avoid having to define every possible status combination. What do you think?

badtst: I like moving the `busy` information to it's dedicated field since I feel it is more closely related to the `capacity` field than to the other status code (we could even imagine a case where the qpu is `PAUSED` **and** `BUSY` where the admins want to drain a huge queue before a maintenance)

I just think that if we provide a helper function `is_accessible()` we must be clear what we consider `accessible` and on what fields it depends on. To me it should rely only on `status` field only. And optional `healthy`/`busy`/`capacity` would be additional information for the  scheduler/user to decide whether they actually want to send a job knowing the current operational details of the qpu.

So in that case `is_accessible` would translate to:
```suggestion
    /// Returns whether a job can currently be submitted to the resource.
    /// Matches the previous behavior of `is_accessible()`
    /// (true when Online).
    pub fn is_accessible(&self) -> bool {
        matches!(
            self.status,
            ResourceStatusCode::Online
        )
    }
}
```

And move the `busy` information to it's dedicated field.

What do you think ? 

@MatthieuMoreau0 @awennersteen

ohtanim: In this case, we should remove 
`/// Matches the previous behavior of `is_accessible()``
because checking online only is incompatible to existing is_accessible().

badtst: Why wouldn't it be ? The current default `status()` implementation directly maps `is_accessible == true -> ResourceStatusCode::Online` and in the case I described above the `ResourceStatusCode` can only have 3 values: `Online|Offline|Paused`

ohtanim: You're right. Yes. you proposed to remove Busy from status codes.

ohtanim: Removed `ResourceStatus::is_accessible()`. User will know from `matches!(status, ResourceStatusCode::Online)`

ohtanim:
@awennersteen @badtst @MatthieuMoreau0 @yoonho @OkuyanBoga I have updated the code to incorporate review comments. Could you please review and approve?

Summary of changes:
- Removed `ResourceStatus::is_accessible()` - User can know by `match(status, ResourceStatusCode::Online)`.
- Removed `ResourceStatusCode::Busy` and add `busy` field to `ResourceStatus`.
- Implemented `status()` for IBM Quantum Compute Service, IBM Quantum System and IQM Server.
   - I noticed that get_gc_v1() was missing from the current dependencies code, so I added for this functionality.

ohtanim:
IQM response:

```
{
  "status": "online",
  "status_reason": "healthy updated at 2026-09-20 12:31:30.062438 +00:00",
  "healthy": true,
  "busy": null,
  "capacity": null,
  "pending_job_count": 0
}
```

IBM Quantum Compute Service response:
```
{
  "status": "online",
  "status_reason": "available",
  "healthy": null,
  "busy": false,
  "capacity": null,
  "pending_job_count": 1
}
```

IBM Quantum System response:
```
{
  "status": "online",
  "status_reason": null,
  "healthy": null,
  "busy": false,
  "capacity": {
    "available_slots": 2,
    "max_slots": 10
  },
  "pending_job_count": null
}
```

lua/lua_qrmi.c
```diff
@@ -761,6 +779,105 @@ static int l_metadata(lua_State *L) {
...
+        lua_pushinteger(L, (lua_Integer)capacity->available_slots);
+        lua_setfield(L, -2, "available_slots");
+        lua_pushinteger(L, (lua_Integer)capacity->max_slots);
+        lua_setfield(L, -2, "max_slots");
+        qrmi_resource_capacity_free(capacity);
+    } else {
+        lua_pushnil(L);
+    }
+    lua_setfield(L, -2, "capacity");
+
```
awennersteen:
```suggestion
    qrmi_resource_status_free(status)
```
?

Seems that we only free this when not successfull

src/ibm/quantum_compute_service.rs
```diff
@@ -343,9 +480,8 @@ impl QuantumResource for IBMQuantumComputeService {
         if status == models::job_response::Status::Running
             || status == models::job_response::Status::Queued
         {
-            jobs_api::cancel_job_jid(&self.config, task_id, None, None)
-                .await
-                .map_err(|e| classify(e, ResourceKind::Job))?;
+            let _ = jobs_api::cancel_job_jid(&self.config, task_id, None, None).await;
+            //jobs_api::delete_job_jid(&self.config, task_id, None).await?;
         }
```
awennersteen: I note two things here for you:

1. you don't check the return code of the cancel request?
2. the commented-out line 484?

examples/qrmi/python/pasqal_cloud/example.py
```diff
@@ -34,7 +34,7 @@
 print(f"Selected resource: id={qrmi.resource_id()} type={str(qrmi.resource_type())}")
 
 # Check if QR it's accessible
-is_avail = qrmi.is_accessible()
+is_avail = qrmi.status().status == ResourceStatusCode.Online
```
awennersteen: How come this supposedly works for us (I guess un-tested), but is a `json.dumps(qrmi.status())` for others?

Is this the intended way for it to work?

awennersteen:
I will run a local test now. But I think this is all good for the general code. (although it looks like Thomas has already reviewed so it's probably OK)


For simplicity, I'm thinking to open a different PR to fix any issues with Pasqal code related to this PR.


So I'm in principle OK to merge if you're OK with it. but let's hold off on the release a little bit (24h), to ensure we can align all the vendor implementations?

awennersteen (approved):
lgtm

examples/qrmi/python/pasqal_cloud/example.py
ohtanim: I'm following the original implementation.

In the Pasqal example, the result is stored in `is_avail`, so I updated that part to use the new code.
In the other examples, I replaced

`print(qrmi.is_accessible())`

with a `json.dumps(...)` call before printing the result.

src/ibm/quantum_compute_service.rs
ohtanim: Thanks, I was not sure why old code is here..

examples/qrmi/python/pasqal_cloud/example.py
awennersteen: ah, gotcha thanks! Sorry, I should have realized!

src/ibm/quantum_compute_service.rs
ohtanim: Fixed (reverted)

ohtanim:
> So I'm in principle OK to merge if you're OK with it. but let's hold off on the release a little bit (24h), to ensure we can align all the vendor implementations?

Sure. I wasn't planning to merge it until all vendors had finished reviewing it either.

@machjn Could you please review this PR for Alice & Bob implementation ?

awennersteen:
Confirmed against Pasqal infra

lua/lua_qrmi.c
```diff
@@ -761,6 +779,105 @@ static int l_metadata(lua_State *L) {
...
+        lua_pushinteger(L, (lua_Integer)capacity->available_slots);
+        lua_setfield(L, -2, "available_slots");
+        lua_pushinteger(L, (lua_Integer)capacity->max_slots);
+        lua_setfield(L, -2, "max_slots");
+        qrmi_resource_capacity_free(capacity);
+    } else {
+        lua_pushnil(L);
+    }
+    lua_setfield(L, -2, "capacity");
+    qrmi_resource_status_free(status)
```
badtst:
```suggestion
    qrmi_resource_status_free(status);
```

lua/lua_qrmi.c
```diff
@@ -761,6 +779,105 @@ static int l_metadata(lua_State *L) {
     return 1;
 }
 
+/**
+ * @brief `resource:status()` - Fetch detailed status information as a Lua table.
+ *
+ * Wraps qrmi_resource_status() and its field accessors
+ * (qrmi_resource_status_code(), qrmi_resource_status_reason(),
+ * qrmi_resource_status_healthy(), qrmi_resource_status_pending_job_count(),
+ * qrmi_resource_status_capacity(), qrmi_resource_status_is_accessible()),
```
badtst:
```suggestion
 * qrmi_resource_status_capacity()),
```

badtst (changes requested):
Missing semicolon in the lua bindings. Otherwise lgtm !

lua/lua_qrmi.c
ohtanim: Thanks. Fixed.

lua/lua_qrmi.c
ohtanim: Thanks. Fixed.

src/lib.rs
```diff
@@ -147,8 +148,81 @@ pub trait QuantumResource: Send + Sync {
     ///     Ok(())
     /// }
     /// ```
+    #[deprecated(
+        since = "0.25.0",
+        note = "use `status().await?.is_accessible()` instead"
```
badtst: This one slipped through !

examples/qrmi/c/resource_providers/src/providers.c
```diff
@@ -73,17 +73,28 @@ int main(int argc, char *argv[]) {
...
+      rc = qrmi_resource_status(resources.resources[i], &res_status);
       if (rc == QRMI_RETURN_CODE_SUCCESS) {
-        if (is_accessible) {
-          fprintf(stdout, "%s can be accessed.\n", id);
+        QrmiResourceStatusCode status;
+        rc = qrmi_resource_status_code(res_status, &status);
+        qrmi_resource_status_free(res_status);
+        if (rc == QRMI_RETURN_CODE_SUCCESS) {
+          if (status == QRMI_RESOURCE_STATUS_CODE_ONLINE) {
+            fprintf(stderr, "%s cannot be accessed.\n", id);
```
badtst: In that case, the resource is accessible right ?

src/cext.rs
```diff
@@ -862,24 +871,465 @@ pub unsafe extern "C" fn qrmi_resource_is_accessible(
...
+        return ReturnCode::NullPointerError;
+    }
+    unsafe {
+        let _ = Box::from_raw(ptr);
+    };
+    ReturnCode::Success
+}
+
+/// @ingroup QrmiResourceStatus
+/// Returns the resource's status code (Online/Offline/Paused/Busy).
```
badtst: No more busy status code here also

lua/lua_qrmi.c
```diff
@@ -126,6 +126,17 @@ static const char *task_status_to_string(QrmiTaskStatus s) {
     }
 }
 
+/**
+ * @brief Convert a QrmiResourceStatusCode enum value to its Lua-facing string form.
+ *
+ * @param s A QrmiResourceStatusCode value as returned by qrmi_resource_status_code().
+ * @return One of "online", "offline", "paused", "busy", or "unknown" if
```
badtst: Outdated busy reference

badtst (approved):
lgtm thanks !

src/lib.rs
ohtanim: fixed. thanks,

examples/qrmi/c/resource_providers/src/providers.c
ohtanim: fixed. good catch.

lua/lua_qrmi.c
ohtanim: fixed.

src/cext.rs
ohtanim: fixed.

PR #250, opened 2026-09-14
Fix IBM Runtime task cancellation for empty cancel requests

### LINKED ISSUE #249: [Bug]: IBM Runtime task cancellation can fail for empty cancel requests

1. Configure an IBM Runtime quantum resource in QRMI.
2. Start a quantum task using the IBM Runtime resource and obtain the remote task ID.
3. While the remote task is still active, call `task_stop()` for that task.
4. Observe the cancellation request sent to the IBM Runtime cancel endpoint.
5. The empty POST cancellation request can be rejected when it is sent without an explicit `Content-Length: 0` header.

### What happens?

IBM Runtime task cancellation can fail even though the cancel endpoint does not require a request body.

The generated empty POST request does not explicitly include `Content-Length: 0`. During testing, the cancellation request was rejected. Adding `Content-Length: 0` allowed the cancellation request to be accepted and the remote task to be cancelled.

Additionally, the IBM Runtime `task_stop()` implementation discards the result of the cancellation API call. As a result, cancellation failures can be hidden from the caller instead of being propagated as QRMI errors.

Expected behavior:

- Empty cancellation POST requests should be correctly framed with `Content-Length: 0`.
- Errors returned by the cancellation API should be propagated to the caller.
- `task_stop()` should successfully cancel the remote task when the provider accepts the request.

Calling task_stop() for an active IBM Runtime task should send a valid cancellation request to the provider and cancel the corresponding remote task.

If the provider rejects the cancellation request or another API error occurs, QRMI should propagate that error to the caller rather than silently ignoring it.

```shell
During end-to-end testing, the IBM Runtime cancellation request reached the cancel endpoint but was rejected.

The cancellation path succeeded after explicitly adding:

Content-Length: 0

to the empty POST request.

Before the fix:
task_stop() -> cancel API request -> request rejected

After the fix:
task_stop() -> cancel API request with Content-Length: 0 -> cancellation accepted -> remote task Cancelled
```

### Additional context

The issue was discovered while testing remote task lifecycle management through QRMI.

Two related behaviors were identified:

1. The generated IBM Runtime cancel request did not explicitly set `Content-Length: 0` for the empty POST request.
2. The result of the cancellation API call in `task_stop()` was discarded, which could hide cancellation failures from callers.

A minimal fix has been implemented and validated locally:

- add `Content-Length: 0` to the cancel request,
- propagate cancellation API errors from `task_stop()`.

Validation completed successfully with:

- `cargo fmt --all -- --check`
- `cargo build`
- `cargo test`
- 33 unit tests passed
- 16 doc tests passed

The cancellation behavior was also validated end-to-end against a real IBM Quantum workload.

konepalli:
## Description of Change

Fix IBM Runtime task cancellation when the cancel endpoint is invoked with an empty POST request.

The generated cancellation request did not explicitly send `Content-Length: 0`, which could cause the cancellation request to be rejected.

This change:

- adds `Content-Length: 0` to the IBM Runtime cancel request;
- propagates cancellation API errors from `task_stop()` instead of discarding them.

The fix was also validated end-to-end against a real IBM Quantum workload.

## Checklist ✅

- [ ] Have you updated the relevant documentation to reflect this change? (Not required; no user-facing configuration/API change)

## Ticket

- [x] Fixes #249
- [ ] Is Part of #


konepalli:
This fix was validated locally with cargo fmt --all -- --check, cargo build, and cargo test (33 unit tests and 16 doc tests passed). It was also validated against a real IBM Quantum Runtime cancellation flow, where the cancellation request succeeded with the corrected empty-request framing

konepalli:
The lychee-checker failure appears unrelated to this PR. It is reporting HTTP 403 for the existing Zenodo badge URL in README.md. This PR only modifies the IBM Quantum Compute client cancellation request and QRMI cancellation error propagation; it does not modify README.md or documentation links

ohtanim (approved):
LGTM. 

but it is a bit strange to me. we didn't touch this code for a long time, and have been using same dependencies with Cargo.lock. It was working correctly until recently, and I was able to cancel it without any issues. I will hear the comment from server side development.

ohtanim:
> The lychee-checker failure appears unrelated to this PR. It is reporting HTTP 403 for the existing Zenodo badge URL in README.md. This PR only modifies the IBM Quantum Compute client cancellation request and QRMI cancellation error propagation; it does not modify README.md or documentation links

Yes, this is not related to your changes. Since last weekend, Zenodo URL access has been frequestly down.

PR #253, opened 2026-09-15
Propagate IBM Quantum Compute cancellation errors

### LINKED ISSUE #252: [Bug]: IBMQuantumComputeService task_stop silently ignores cancellation errors

Configure an IBMQuantumComputeService resource and submit a task.
Ensure the task reaches the Queued or Running state.
Call task_stop(task_id) while the task is active.
Have the provider-side POST /jobs/{id}/cancel request return an error.
Observe the return value from task_stop().

In the current implementation, the result of the cancellation request is explicitly discarded:

let _ = jobs_api::cancel_job_jid(&self.config, task_id, None, None).await;

### What happens?

IBMQuantumComputeService::task_stop() silently ignores errors returned by cancel_job_jid().

If the provider-side cancellation request fails, task_stop() still returns Ok(()). This can make the caller believe that the stop operation succeeded even though the remote task was not successfully cancelled.

If cancel_job_jid() fails, task_stop() should return an error to the caller instead of reporting success.

The cancellation error should be classified using the existing QRMI error handling mechanism:

jobs_api::cancel_job_jid(&self.config, task_id, None, None)
    .await
    .map_err(|e| classify(e, ResourceKind::Job))?;

This is also consistent with the existing behavior of IBMQiskitRuntimeService::task_stop().

```shell
The problem is visible directly in the current IBMQuantumComputeService::task_stop() implementation.

Current implementation:

let _ = jobs_api::cancel_job_jid(&self.config, task_id, None, None).await;

The returned Result is discarded, after which task_stop() returns Ok(()).
```

### Additional context

IBMQiskitRuntimeService::task_stop() already handles the same cancellation operation by classifying and propagating the error:

jobs_api::cancel_job_jid(&self.config, task_id, None, None)
    .await
    .map_err(|e| classify(e, ResourceKind::Job))?;

Applying the same error-handling behavior to IBMQuantumComputeService would make the two IBM implementations consistent and ensure that callers can reliably determine whether a remote cancellation succeeded.

I have prepared a minimal change implementing this behavior and validated it with:

cargo fmt --check
cargo check
cargo test — 33 unit tests passed
16 doc tests passed
git diff --check

konepalli:
## Description of Change

`IBMQuantumComputeService::task_stop()` previously discarded errors returned by `cancel_job_jid()`, which could cause `task_stop()` to return `Ok(())` even when the provider-side cancellation request failed.

This change classifies and propagates cancellation errors using the existing `classify(..., ResourceKind::Job)` mechanism, consistent with `IBMQiskitRuntimeService::task_stop()`.

Validation completed:
- `cargo fmt --check`
- `cargo check`
- `cargo test` — 33 unit tests passed and 16 doc tests passed
- `git diff --check`

## Checklist ✅

## Ticket
- [x] Fixes #252
- [ ] Is Part of #


ohtanim (approved):
LGTM. Thanks

PR #255, opened 2026-09-15
Correct IBM task stop documentation

### LINKED ISSUE #254: [Bug]: IBM task_stop documentation incorrectly states that jobs are deleted

1. Inspect `task_stop()` in:
   - `src/ibm/quantum_compute_service.rs`
   - `src/ibm/qiskit_runtime_service.rs`
2. Compare the method documentation with the implementation.
3. The documentation states that an active job is cancelled and then deleted using `DELETE /jobs/{id}`.
4. The implementation only sends `POST /jobs/{id}/cancel` for jobs in the `Running` or `Queued` state. No delete request is performed.

### What happens?

The documentation for both IBM implementations does not match the actual `task_stop()` behavior.

It currently states that the job is cancelled before being deleted with `DELETE /jobs/{id}`. However, the implementation only cancels jobs that are `Running` or `Queued`.

There is also a commented-out `delete_job_jid()` call remaining in both implementations.

The documentation should accurately describe the current behavior of `task_stop()`:

- The job status is checked.
- If the job is `Running` or `Queued`, a cancellation request is sent using `POST /jobs/{id}/cancel`.
- The job is not deleted.

The stale commented-out `delete_job_jid()` call should also be removed.

```shell
No runtime error is involved. This is a documentation/implementation discrepancy visible directly in the source.
```

### Additional context

The generic `QuantumResource::task_stop()` API describes stopping a task and does not require provider-side deletion.

The current IBM implementations already implement cancellation behavior. This change is intended only to align the documentation with the existing behavior and remove stale commented-out code; it does not change runtime behavior.

konepalli:
## Description of Change

The documentation for `task_stop()` in `IBMQuantumComputeService` and `IBMQiskitRuntimeService` stated that active jobs are cancelled and then deleted using `DELETE /jobs/{id}`.

The implementations only cancel jobs that are `Running` or `Queued`; they do not delete the provider-side job.

This change:
- updates the documentation to describe the existing cancellation behavior accurately;
- clarifies that both `Running` and `Queued` jobs can be stopped;
- removes the stale commented-out `delete_job_jid()` calls.

This is a documentation and cleanup change only and does not modify runtime cancellation behavior.

Validation completed:
- `cargo fmt --check`
- `cargo check`
- `cargo test` — 33 unit tests passed and 16 doc tests passed
- `git diff --check`

## Checklist ✅

## Ticket
- [x] Fixes #254
- [ ] Is Part of #


ohtanim (approved):
LGTM.

